PR #24, opened 2016-09-21
Add support for Devtools

mukundgoel:
**Summary:**
Integrating Redux devtools with the Electrode sample app to allow users to easily debug their application issues. There is not much data in the store given it is a sample application, but it will serve as a reference on integrating devtools with the Electrode app.

**Screenshot:**
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ npm install
 $ NODE_ENV=development gulp hot
 ```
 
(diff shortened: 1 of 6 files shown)


caoyangs:
Thank you for your PR. Could you update README as well?

mukundgoel:
added @caoyangs

PR #29, opened 2016-10-13
changed combine reducers to fix Cannot read property 'data' of undefined

ibulmer:
See issue 28.


ibulmer:
Had some linting errors. Will fix and resubmit.

Issue #28, opened 2016-10-10
home.jsx Cannot read property 'data' of undefined


ibulmer:
Steps taken:

1. git clone https://github.com/electrode-io/electrode-boilerplate-universal-react-node.git
2. cd electrode-boilerplate-universal-react-node
3. npm install
4. app.jsx 6: deleted import { createHistory } from "history";
5. app.jsx 29: deleted history={createHistory()
6. In the app clicked on SSR Caching Simple Type Example link.

Getting the following error: home.jsx Cannot read property 'data' of undefined

mtbarta:
you can still provide a history object. See below.

```javascript
import { Router, browserHistory } from "react-router";
...
window.webappStart = () => {
  Resolver.render(
    () =>
      <Provider store={store}>
        <div>
          <Router history={browserHistory}>{ routes }</Router>
          <DevTools />
        </div>
      </Provider>,
    document.querySelector(".js-content")
  );
};
```

ibulmer:
The comment from mtbarta was helpful in solving a lot of problems I was having. On this particular issue I believe what is causing the problem is the way the root reducer is being defined in client/app.jsx.  In the app when you navigate to any other page, the parameter s in the root-reducer function is undefined and that is what is causing this error. If you directly go to the page in your browser without navigating to it the issue does not happen.  

I have always seen the root reducer set up something like:
```
const rootReducer = combineReducers({
  data: dataReducer,
  count: countReducer
})
```
I made that change and it has fixed the issue. Can submit a pr to show what I did if it would be helpful.

ananavati:
@ibulmer thanks for reporting and fixing the issue, closing this issue since the bug fix was merged.

PR #30, opened 2016-10-13
home.jsx Cannot read property 'data' of undefined

ibulmer:
See issue #28


ananavati:
thanks @ibulmer for fixing this error. hope you are enjoying electrode.

ibulmer:
Yes I like it a lot :)

PR #40, opened 2016-11-08
Implement basic PWA features

Aweary:
Implements basic PWA features. See https://github.com/electrode-io/electrode-archetype-react-app/pull/46 for more details.

This PR includes the required `manfiest.json` and generated assets required for add to home functionality. It also makes some changes to `webapp` so that it properly parses and loads both the manifest file and service worker.


client/manifest.json
```diff
@@ -0,0 +1,18 @@
+{
+	"name": "Electrode Boilerplate",
+	"icons": [
+		{
+			"src": "./icons/android-chrome-192x192.png",
```
Aweary: Thanks to the webpack plugin we're using to load the manifest file, we can just specify the paths like we would if we're requiring them, and it will bundle them + replace with the new path for us.

server/plugins/webapp/register-sw.html
```diff
@@ -0,0 +1,7 @@
+<script>
```
Aweary: It could be, but it's such a small file I don't think it makes sense to incur the cost of a network request to load it, so I inlined it.

Aweary:
@ananavati add to home works as well

screen shot 2016-11-08 at 1 28 30 pm

The one caveat that we're working out is that the `sw.js` file needs to be generated first, so for `gulp dev` you need to run `gulp build` first to generate the file. After discussing this @donsalari is working on using a gulp script instead of the webpack plugin, and that will run before `build` and `dev`.

ananavati:
gitignore https://github.com/electrode-io/electrode-boilerplate-universal-react-node/pull/40/files#diff-8e581f12eb6103b4945e19c3749fe66d

server/plugins/webapp/register-sw.html
ananavati: for now it makes sense to be inline but before going to production this piece of code will need to handle all cases.

1. What if SW registration fails ?
2. Showing a message like `Caching complete! Future visits will work offline.` when SW is in state  `installed`.

jmcriffey:
LGTM 👍

PR #48, opened 2016-12-02
PWA features / Getting started with push notifications.

donsalari:
PWA Features : https://github.com/donsalari/electrode-boilerplate-universal-react-node/blob/master/README.md#progressive-web-app-pwa-features-supported-by-the-electrode-framework

Push Notifications: https://github.com/donsalari/electrode-boilerplate-universal-react-node/blob/master/README.md#instructions-for-setting-up-push-notification
and 
https://github.com/donsalari/electrode-boilerplate-universal-react-node/blob/master/README.md#instructions-for-sending-a-push-notification

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,14 @@ $ gulp server-prod
 
 - Running in the selected environment should load the appropriate configuration settings
 
+- Start the electrode app with `service workers`
+
+```bash
+$ gulp build
+$ gulp server
+```
+Service worker currently does not work with webpack dev server. You need to build first and then run the server.
+
 ## Instructions for boostrapping new electrode application
 You can bootstrap a new electrode webapplication from scratch by doing:
 
@@ -52,10 +60,10 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
 ## Progressive Web App (PWA) features supported by the Electrode framework  
 #### 1. Offline first  
   Offline first lets your app run without a network connection. At the same time it provides a great performance boost for repeat visit to your web site.
-  This is done with a service worker and by pre-caching your static assets as well as runtime caching of React routes.  
+  This is done with a service worker and by pre-caching your static assets as well as runtime caching of dynamic server routes and external resources.  
   [Learn More](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#2-cache)  
 #### 2. Add To Homescreen  
-  After visiting you website, users will get a prompt (if the user has visited your site at least twice, with at least five minutes between visits.) to add your application to their homescreen (web or [mobile](https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android)). Combined with offline caching, this means your web app can be used exactly like a native application.  
+  After visiting your website, users will get a prompt (if the user has visited your site at least twice, with at least five minutes between visits.) to add your application to their homescreen (web or [mobile](https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android)). Combined with offline caching, this means your web app can be used exactly like a native application.  
   [Learn More](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#1-manifest)  
 #### 3. Push Notifications  
   Web push notifications allow users to opt-in to timely updates from sites they love and allow you to effectively re-engage them with customized, relevant content.  
@@ -107,7 +115,11 @@ if ("serviceWorker" in navigator) {
 The service worker is ready to accept `push` from the server. On receiving the push, it will provide the `notification` to the browser.
 
 #### 4. Setting up your API_KEY and GCM_SENDER_ID  
-We will be needing the API_KEY and GCM_ENDPOINT to send the messages from the server. To generate these values, visit [firebase](https://console.firebase.google.com) and create a new project. Click on the setting icons and open `Project settings`. Navigate to the `CLOUD MESSAGING` tab to view your `Server key or Legacy Server key` (API_KEY) and the `Sender ID`. You need to update your `manifest` in `sw-config.js` to update the (gcm_sender_id)[https://github.com/electrode-io/electrode-boilerplate-universal-react-node/blob/master/config/sw-config.js#L18].  
+We will be needing the API_KEY and GCM_ENDPOINT to send the messages from the server.  
+To generate these values, visit [Firebase](https://console.firebase.google.com) and create a new project.  
+Click on the setting icons and open `Project settings`.  
+Navigate to the `CLOUD MESSAGING` tab to view your `Server key or Legacy Server key` (API_KEY) and the `Sender ID`.  
+You need to update your `manifest` in `sw-config.js` to update the [gcm_sender_id](https://github.com/electrode-io/electrode-boilerplate-universal-react-node/blob/master/config/sw-config.js#L18).  
 
 ## Instructions for sending a push notification  
 Now that we have our service worker up and running, we can send a `push` with the following steps:  
--- a/README.md
+++ b/README.md
@@ -49,12 +49,24 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
   - [Electrode Confippet](https://github.com/electrode-io/electrode-confippet)
   - [Electrode Javascript Bundle Viewer](https://github.com/electrode-io/electrify)
 
+## Progressive Web App (PWA) features supported by the Electrode framework  
+#### 1. Offline first  
+  Offline first lets your app run without a network connection. At the same time it provides a great performance boost for repeat visit to your web site.
+  This is done with a service worker and by pre-caching your static assets as well as run time caching of react routes.  
+  [Learn More](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#2-cache)  
+#### 2. Add To Homescreen  
+  After visiting you website, users will get a prompt to add your application to their homescreen (web or mobile). Combined with offline caching, this means your web app can be used exactly like a native application.  
+  [Learn More](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#1-manifest)  
+#### 3. Push Notifications  
+  Web push notifications allow users to opt-in to timely updates from sites they love and allow you to effectively re-engage them with customized, relevant content.  
+  We will learn about Push Notifications in the next couple of sections.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -49,6 +49,100 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
   - [Electrode Confippet](https://github.com/electrode-io/electrode-confippet)
   - [Electrode Javascript Bundle Viewer](https://github.com/electrode-io/electrify)
 
+## Progressive Web App (PWA) features supported by the Electrode framework  
+#### 1. Offline first  
+  Offline first lets your app run without a network connection. At the same time it provides a great performance boost for repeat visit to your web site.
+  This is done with a service worker and by pre-caching your static assets as well as run time caching of react routes.  
+  [Learn More](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#2-cache)  
+#### 2. Add To Homescreen  
+  After visiting you website, users will get a prompt to add your application to their homescreen (web or mobile). Combined with offline caching, this means your web app can be used exactly like a native application.
```
Aweary: The "Add to Homescreen" prompt heuristics are actually a little more complicated, and Chrome for Android will actually prompt the user to do it AFAIK. You can see them listed [here](https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android), but it's an ever-changing list of requirements.

We take care of most of those, but one that app authors should be aware of is:

> The user has visited your site at least twice, with at least five minutes between visits.

So they won't see the banner immediately on their first visit. Another thing to note is that users can still manually add to home from their mobile browser menu.

README.md
```diff
@@ -49,6 +49,100 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
   - [Electrode Confippet](https://github.com/electrode-io/electrode-confippet)
   - [Electrode Javascript Bundle Viewer](https://github.com/electrode-io/electrify)
 
+## Progressive Web App (PWA) features supported by the Electrode framework  
+#### 1. Offline first  
+  Offline first lets your app run without a network connection. At the same time it provides a great performance boost for repeat visit to your web site.
+  This is done with a service worker and by pre-caching your static assets as well as run time caching of react routes.
```
Aweary: > run time caching of react routes.  

=>

> runtime caching of React routes.

README.md
```diff
@@ -49,6 +49,100 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
...
+  [Learn More](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#2-cache)  
+#### 2. Add To Homescreen  
+  After visiting you website, users will get a prompt to add your application to their homescreen (web or mobile). Combined with offline caching, this means your web app can be used exactly like a native application.  
+  [Learn More](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#1-manifest)  
+#### 3. Push Notifications  
+  Web push notifications allow users to opt-in to timely updates from sites they love and allow you to effectively re-engage them with customized, relevant content.  
+  We will learn about Push Notifications in the next couple of sections.
+
+## Instructions for setting up push notification
+Push Notification is based on service workers because service workers operate in the background. So we need to register our service worker first.
```
Aweary: > Push Notification is based on service workers because service workers operate in the background

I would rephrase to:

> The [Push API](https://developer.mozilla.org/en-US/docs/Web/API/Push_API) requires a registered service worker so it can send notifications in the background when the web application isn't running.

Aweary: Any maybe add a link to the relavant Codelab from Google for those interested in more details

> Check out the [Adding Push Notifications to a Web App](https://developers.google.com/web/fundamentals/getting-started/codelabs/push-notifications/) Codelab provided by Google for an in-depth guide on how push notifications and service workers work together.

README.md
```diff
@@ -49,6 +49,100 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
...
+  [Learn More](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#1-manifest)  
+#### 3. Push Notifications  
+  Web push notifications allow users to opt-in to timely updates from sites they love and allow you to effectively re-engage them with customized, relevant content.  
+  We will learn about Push Notifications in the next couple of sections.
+
+## Instructions for setting up push notification
+Push Notification is based on service workers because service workers operate in the background. So we need to register our service worker first.  
+Check out [this guideline](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#how-do-i-generate-a-manifestjson-and-a-service-worker-file) to generate a service worker in an electrode app.  
+
+Next we need to add a `push` event to this existing service worker:
```
Aweary: We should clarify that you only need to listen to the `push` event if you're going to be sending notifications to the client from a server.

README.md
```diff
@@ -49,6 +49,100 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
...
+#### 3. Push Notifications  
+  Web push notifications allow users to opt-in to timely updates from sites they love and allow you to effectively re-engage them with customized, relevant content.  
+  We will learn about Push Notifications in the next couple of sections.
+
+## Instructions for setting up push notification
+Push Notification is based on service workers because service workers operate in the background. So we need to register our service worker first.  
+Check out [this guideline](https://github.com/electrode-io/electrode-archetype-react-app/blob/master/README.md#how-do-i-generate-a-manifestjson-and-a-service-worker-file) to generate a service worker in an electrode app.  
+
+Next we need to add a `push` event to this existing service worker:  
+#### 1. Create a new file `sw.js`,
```
Aweary: Maybe add more detail about what this is doing (also it doesn't need to be called `sw.js`). Something like:

> In order to respond to push notifications events received from a remote server we need to listen for the `push` event on the active service worker. Since our service worker file is generated automatically we need to use the `importScripts` API, which lets us execute additional scripts in the context of the service worker.

README.md
```diff
@@ -49,6 +49,100 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
...
+module.exports = {
+  cache: {
+    importScripts: ['./sw.js']
+    }
+}
+```
+[Sample file](https://github.com/electrode-io/electrode-boilerplate-universal-react-node/blob/master/config/sw-config.js)
+
+
+#### 3. Now we can register this `push` event in the browser and install the new service worker:
```
Aweary: This step needs to be done regardless of whether push notifications are used. Maybe it should be part of a guide on registration service workers, and then we can link them there.

README.md
```diff
@@ -49,6 +49,100 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
...
+      });
+    });
+});
+```
+[Sample subscription](https://github.com/electrode-io/electrode-boilerplate-universal-react-node/blob/master/client/components/push-notifications.jsx#L73-L86)  
+
+Typically, after the user subscribes, we send the subscription information to the server and the server uses the `subscriptionId` to trigger a notification.
+
+#### 2. Sending Messages  
+Sending message is as easy as executing a curl command. The curl command contains the `subscriptionId`, `API_KEY` (Your cloud messaging API key from [firebase](https://console.firebase.google.com)) and [GCM_ENDPOINT](https://github.com/electrode-io/electrode-boilerplate-universal-react-node/blob/master/client/components/push-notifications.jsx#L145-L146)
```
Aweary: Maybe include details on getting a `GCM` ID in a section above that talks about setting up push notifications from remote servers.

They'll also need to include the `gcm_sender_id` in the `manifest` config: https://github.com/electrode-io/electrode-boilerplate-universal-react-node/blob/master/config/sw-config.js#L18

README.md
```diff
@@ -49,6 +49,100 @@ This will set up an Electrode webapplication which will have 2 of the above 6 mo
...
+
+Alternatively, you can also send notifications from the service worker:
+```
+sendNotification() {
+   const title = 'This Notification'
+   const body = 'Is brought to you by your server worker!'
+   const options = {body};
+   navigator.serviceWorker.ready.then((registration) => {
```
Aweary: Might be worth explaining that `navigator.serviceWorker.ready` is a Promise that will resolve once a service worker is registered, and it returns a reference to the active [ServiceWorkerRegistration](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration)

Aweary:
cc @ananavati @jmcriffey

ananavati:
like the idea of including links to codelab for examples along with our electrode examples.

ananavati:
@donsalari @Aweary can one of you take care of https://travis-ci.org/electrode-io/electrode-boilerplate-universal-react-node/builds/180583425#L2131

README.md
donsalari: 👍

README.md
donsalari: Our service worker registrations works with the file generated with sw-config. I was under the impression that we are doing the re registration to add the push event?

Aweary: Nope, `sw-precache` will include the script in the generated `sw.js` using `importScripts` so we only need to register once

README.md
jmcriffey: If you had an offline game or date/time keeping app, you could send a push notification based on a user event or time instead of the server, right?

README.md
jmcriffey: Maybe make this more general. In practice we will be using runtime caching for CDN assets and API calls on top of React routes:

> runtime caching of routes and external resources

README.md
jmcriffey: That makes sense. I was thinking about generate notifications and not the `push` part.

README.md
donsalari: done

client/components/push-notifications.jsx
```diff
@@ -171,7 +171,7 @@ export default class PushNotifications extends React.Component {
 
       <h3>Push Notification Subscription</h3>
       <code>
-        {JSON.stringify(this.state.subscription, null, 4)}
+        {JSON.stringify(this.state.subscription)}
```
donsalari: Nope. It generates the curl command. You would mostly copy paste it in the terminal to send a push notification.

jmcriffey: Cool, looks good then.

Aweary: This string is actually rendered to the screen, the curl command is also rendered below it. It wasn't very pretty though, it might be better to just do something like:

```jsx
<h3> Endpoint </h3>
<code>
{this.state.subscription.endpoint}
</code>
```

PR #49, opened 2016-12-06
Implement critical CSS injection

Aweary:
Moves some inline styles to external files to demonstrate critical CSS inlining. Adds a new marker to `webapp` that injects a style tag with the parsed critical css.

### DIFF
--- a/server/plugins/webapp/index.js
+++ b/server/plugins/webapp/index.js
@@ -50,6 +50,12 @@ function getIconStats(iconStatsPath) {
   return iconStats;
 }
 
+function getCriticalCSS(path) {
+  const criticalCSSPath = Path.resolve(process.cwd(), path);
+  const criticalCSS = fs.readFileSync(criticalCSSPath).toString();
+  return `<style>${criticalCSS}</style>`;
+}
+
 function makeRouteHandler(options, userContent) {
   const CONTENT_MARKER = "{{SSR_CONTENT}}";
   const HEADER_BUNDLE_MARKER = "{{WEBAPP_HEADER_BUNDLES}}";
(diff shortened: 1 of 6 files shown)


Aweary:
Requires changes made in https://github.com/electrode-io/electrode-archetype-react-app/pull/56

server/plugins/webapp/index.js
```diff
@@ -9,6 +9,14 @@ const assert = require("assert");
 const HTTP_ERROR_500 = 500;
 const HTTP_REDIRECT = 302;
 
+const CONTENT_MARKER = "{{SSR_CONTENT}}";
+const HEADER_BUNDLE_MARKER = "{{WEBAPP_HEADER_BUNDLES}}";
+const BODY_BUNDLE_MARKER = "{{WEBAPP_BODY_BUNDLES}}";
+const TITLE_MARKER = "{{PAGE_TITLE}}";
+const PREFETCH_MARKER = "{{PREFETCH_BUNDLES}}";
+const META_TAGS_MARKER = "{{META_TAGS}}";
+const CRITICAL_CSS_MARKER = "{{CRITICAL_CSS}}";
```
Aweary: Moved these out of makeRouteHandler since we enforce a max statement rule, and there's no reason these need to be defined in makeRouteHandler

server/plugins/webapp/index.js
```diff
@@ -103,14 +116,20 @@ function makeRouteHandler(options, userContent) {
...
+        : "";
       return `${manifestLink}${cssLink}`;
     };
 
     const makeBodyBundles = () => {
       const js = bundleJs();
+      const css = bundleCss();
+      const cssLink = css && criticalCSS
+      ? `<link rel="stylesheet" href="${css}" />`
+      : "";
```
Aweary: We want to load the stylesheet after the body content if there is critical CSS, so it doesn't block rendering

PR #50, opened 2016-12-06
Support multiple entries

jmcriffey:
@ananavati @Aweary 

This PR enables multiple entry points. It basically reads in the assetsByChunk name and matches them to their asset names. I have tested this setup with and without an `entry.config.js` in both production and dev/hot mode.


jmcriffey:
Our next step will be moving this server code to the webapp module and use that instead.

ananavati:
👍

PR #57, opened 2016-12-09
setup initial redux store when SSR is disabled

animesh10:


packages/generator-electrode/generators/webapp/templates/server/plugins/webapp/react-webapp.js
```diff
@@ -72,6 +72,14 @@ function makeRouteHandler(options, userContent) {
     const renderJs = RENDER_JS && mode !== "nojs";
     const renderSs = RENDER_SS && mode !== "noss";
 
+    if (!renderSs) {
```
jchip: I think we should keep the `noss` mode as is.  and add a new mode for this behavior. maybe `datass`

packages/generator-electrode/generators/webapp/templates/server/plugins/webapp/react-webapp.js
```diff
@@ -72,6 +72,14 @@ function makeRouteHandler(options, userContent) {
...
+      if (options.request.server && options.request.server.app) {
+        options.request.app.disableSSR = true;
```
jchip: inconsistent fields.  accessing `.app.disableSSR` but checking for `.server.app` above.

packages/generator-electrode/generators/webapp/templates/server/plugins/webapp/react-webapp.js
```diff
@@ -71,6 +71,13 @@ function makeRouteHandler(options, userContent) {
     const mode = options.mode;
     const renderJs = RENDER_JS && mode !== "nojs";
     const renderSs = RENDER_SS && mode !== "noss";
+    const dataSs = RENDER_SS && mode === "datass";
+
+    if (dataSs) {
+      if (options.request && options.request.app) {
+        options.request.app.disableSSR = true;
+      }
+    }
```
jchip: change the line 73-80 thing to:

```js
let renderSs = RENDER_SS;
if (renderSs) {
  if ( mode === "noss") {
    renderSs = false;
  } else if (mode === "datass") {
    if (options.request && options.request.app) {
      options.request.app.disableSSR = true;
    }
  }
}
```

packages/generator-electrode/generators/webapp/templates/server/plugins/webapp/react-webapp.js
```diff
@@ -147,7 +154,7 @@ function makeRouteHandler(options, userContent) {
       return p.then((c) => renderPage(c));
     };
 
-    return renderSSRContent(renderSs ? userContent : "");
+    return renderSSRContent((renderSs||dataSs) ? userContent : "");
```
jchip: now you can undo this change.

PR #58, opened 2016-12-09
Correct and simplify entry point handling

Aweary:
Companion PR to https://github.com/electrode-io/electrode-react-webapp/pull/12

* Fixes some dependency location issues (`lodash`, `gulp`, `penthouse`)
* Simplifies multi and single entry point applications by always using the chunk name in dev
* Writes stats file to configurable build artifacts folder for consumption by `electrode-react-webapp` in dev
* Fixes an issue where chunks generated by `importScripts` were overwriting the actual app bundle (as they were using the same filename)

cc @ananavati @jmcriffey @donsalari


packages/electrode-archetype-react-app/config/webpack/partial/pwa.js
```diff
@@ -106,14 +129,66 @@ module.exports = function () {
...
+            output: {
+              filename: 'stats.json'
+            }
+          }]
+        })
+      );
+    }
+
     return mergeWebpackConfig(config, {
+      __wmlMultiBundle: typeof entry === 'object',
```
jmcriffey: I'm not sure how we actually want to handle this. We really need to audit how `__wmlMultiBundle` is being used. What happens if this partial is applied after another partial that deals with `__wmlMultiBundle`?

I think to be safe, we'll need to keep this at the beginning of the partial pipeline.

packages/electrode-archetype-react-app/config/webpack/partial/pwa.js
```diff
@@ -37,6 +46,19 @@ function getHashedPath(filepath) {
 }
 
 /**
+ * Takes a file path and returns a webpack-dev ompatible
```
jmcriffey: Missed a c here: `returns a webpack-dev compatible`.

packages/electrode-archetype-react-app/config/webpack/partial/pwa.js
Aweary: > I think to be safe, we'll need to keep this at the beginning of the partial pipeline.

The issue there is that we don't know if we've got multiple entry points until we've parsed `importScripts`. I think we can actually get rid of it all together with a little more work. Most of these changes make it possible to treat a single-entry point build just like we do multi-entry point builds (since Webpack gives us a default `chunkName` of `"main"` if there's only one entry point)

jmcriffey: If we make sure the pwa partial is applied first, then that solves the issue right? All of this code will be executed (including `importScripts`) and then the rest of the partials can be applied with the correct assumptions about multiple bundles.

Aweary: That's one potential solution, but I'm not totally sure yet if that would break anything in the `pwa` partial.  It's probably safe, but I do think that removing the distinction is probably best as far as complexity goes, if we can manage it.

jmcriffey: Sounds good.

PR #64, opened 2016-12-12
Add react-toast-dep

jmcriffey:
We are importing `react-toast-dep` but it isn't a dependency. https://github.com/electrode-io/electrode/blob/master/samples/universal-react-node/client/components/home.jsx#L4 

@ananavati @donsalari

### DIFF
--- a/samples/universal-react-node/package.json
+++ b/samples/universal-react-node/package.json
@@ -36,7 +36,8 @@
     "electrode-react-webapp": "^1.1.0",
     "electrode-redux-router-engine": "^1.0.0",
     "electrode-server": "^1.0.0",
-    "electrode-static-paths": "^1.0.0"
+    "electrode-static-paths": "^1.0.0",
+    "react-notify-toast": "^0.1.3"
   },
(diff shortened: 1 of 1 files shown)


jmcriffey:
@ananavati Do we have plans to enable CI for packages in this repo?

jmcriffey:
It looks like our unit tests are not being run. I'll investigate further.

jchip:
unit test in `samples/universal-react-node` is executed.  https://github.com/electrode-io/electrode/blob/master/gulpfile.js#L95

jmcriffey:
@jchip It looks like the test has an error while setting up and it doesn't report it. I'll send a PR to fix it soon.

jchip:
when running the test in the CI build, the test updates the sample's package.json to point to local copies of the electrode modules, and that is passing: https://github.com/electrode-io/electrode/blob/master/gulpfile.js#L33-L36

when just running the `test-boilerplate` with package.json pointing to the published version of the archethype, the test is failing.  We should publish new versions of these modules when we are ready.

PR #66, opened 2016-12-12
configure socket.io to support webpack reporter hot reloading

georgeweiler:
This update adds server code to electrode generator apps necessary to support webpack-reporter hot reloading. See: https://github.com/electrode-io/electrode-webpack-reporter/pull/13#pullrequestreview-12534730

### DIFF
--- a/packages/generator-electrode/generators/app/templates/server/express-server.js
+++ b/packages/generator-electrode/generators/app/templates/server/express-server.js
@@ -35,12 +35,6 @@ const setRouteHandler = () => new Promise((resolve, reject) => {
   );
 });
 
-const startIOserver = () => new Promise((resolve) => {
-  const io = require("socket.io")(5000); //eslint-disable-line global-require, no-magic-numbers
-  io.on("connection", (socket) => socket.emit("hmr"));
-  resolve();
-});
-
 const startServer = () => new Promise((resolve, reject) => {
   app.listen(defaultConfig.$("connections.default.port"), (err) => {
     if (err) {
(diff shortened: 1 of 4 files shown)


packages/generator-electrode/generators/app/templates/server/express-server.js
```diff
@@ -35,6 +35,12 @@ const setRouteHandler = () => new Promise((resolve, reject) => {
   );
 });
 
+const startIOserver = () => new Promise((resolve) => {
```
jchip: Is a promise needed?  There's nothing async.

packages/generator-electrode/generators/app/templates/server/index.js
```diff
@@ -32,4 +32,13 @@ supports.babelRegister({
 supports.cssModuleHook({
   generateScopedName: "[name]__[local]___[hash:base64:5]"
 });
-require<% if (isHapi) { %>("electrode-server")(config, [staticPathsDecor()])<% } else { %>("./express-server")(config)<% } %>;
+require<% if (isHapi) {
+  %>("electrode-server")(config, [staticPathsDecor()]).then((server) => {
```
jchip: note that electrode-server is only used when developing/testing the reporter.  When the reporter is actually being used by the user, it's part of webpack's dev server and there's no electrode server.

jchip:
The webpack reporter is served by the `express` server [webpack-dev-server](https://webpack.github.io/docs/webpack-dev-server.html) listening at port 2992 and its URL is http://localhost:2992/reporter.  What's the reason for requiring any kind of altercation on the app's server?

georgeweiler:
The webpack reporter has a socket.io-client listener on port 5000 (with webpackReporter label). This PR will cause the app to send a 'hot module reload' event when changes are saved so the reporter can reload. The app's server needs config altercations to connect to the reporter's listener.

jchip:
any kind of events to the reporter should be sent by webpack dev server's express server.

georgeweiler:
@ananavati helped me get this functionality working without editing the generator. Thanks for the input @jchip

PR #69, opened 2016-12-14
Ssr Auto Hydration

animesh10:


packages/electrode-react-webapp/lib/index.js
```diff
@@ -106,11 +106,22 @@ function makeRouteHandler(options, userContent) {
...
     const renderJs = RENDER_JS && mode !== "nojs";
-    const renderSs = RENDER_SS && mode !== "noss";
+    let renderSs = RENDER_SS;
+    if (renderSs) {
+      if (mode === "noss") {
+        renderSs = false;
+      } else if ((mode === "datass" || disableSsrFlag) && request.app) {
+        request.app.disableSSR = true;
+      } else if (request.app && request.app.disableSSR) {
+        request.app.disableSSR = false;
```
jchip: this logic here doesn't make sense. could you explain the reason at flipping `disableSSR` from `true` to `false`?

animesh10: once disableSSR is set in request.app, it is not being flipped back if the disableSSRFlag is no longer true.

jchip: this flag has to be at request scope, not global.

PR #75, opened 2016-12-15
[WIP] Sync internal archetype with electrode.io

exogen:
Work in progress. Add outstanding features from internal archetype:

* CSS splitting
* Make `chai-shallowly` available to tests
* DLL support
* JSON loader
* Config override support from root project
* Config merging with `webpack-config` (why is `webpack-partial` still used *also*? I thought we wanted one or the other /cc @jmcriffey)
* Add `webpack-fail-plugin`

TODO:

* Audit new DLL plugin files for consistency with electrode.io patterns (file locations, using properties on `archetype` like `clientSrcDir`, etc.)
* Lint configs. For some reason there is no lint task for the archetypes now..?

In progress: major outstanding changes from `next` internal archetype:

* Babel `build-lib` task, discontinue `babel-register` usage
* Fix Babel plugin resolution
* Move build configs to `dev` archetype


jmcriffey:
@exogen Only the component archetype was updated to not use partials. We haven't yet gotten back to updating the app archetype.

jmcriffey:
As to why we use both: `webpack-config` allows you to extend configs using a normal webpack config pattern instead of exporting a function in the `webpack-partial` style.

ananavati:
@exogen 

I think we should tackle `babel-register` pr separately after this pr. thoughts?

exogen:
@ananavati Totally agree and is what I've got coming up. It'd be easier to make a huge PR with all the merging & conflict resolution I've already done (merging the `dev` config fix plus `babel-register` in particular is a huge pain), but it's going to be worth it to have smaller PRs with separate concerns, and I'll just re-fix all the conflicts.

packages/electrode-archetype-react-app/config/webpack/partial/fail.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+var archDevRequire = require("@walmart/electrode-archetype-react-app-dev/require");
```
jmcriffey: I think this should be:

```js
var archDevRequire = require("../../archetype").devRequire;
```

exogen: You're right, will fix.

exogen:
Also still on the TODO list here: there were DLL-related script commands in `package.json` that were not adapted to the new Gulpfile approach here. I don't have experience writing Gulp tasks but looking at the existing tasks, it shouldn't be too much work. /cc @jmcriffey

ananavati:
/cc @alexritchey This pr will enable extending of webpack config to use your custom configs

`re tweet: Is there any simple solution to extending configs in the archetypes yet? Trying to add some plugins to my webpack config`

jmcriffey:
@ananavati @exogen FYI, I have found for this to work for internal apps, we have to make a couple of small changes to the internal webapp.

packages/electrode-archetype-react-app/config/webpack/webpack.config.hot.js
```diff
@@ -18,6 +20,9 @@ var config = module.exports = _.flow(
   hotConfig()
 )();
 
+
+config.devServer = {}; // use webpack default verbosity
```
jmcriffey: This actually causes hot mode to not work. The internal and external archetypes are a bit different in how they run their dev servers. It looks like removing this line fixes the issue. We might need to dive into the exact params to get "default verbosity" if it is important.

ananavati:
closing this pr, changes merged as part of #109

PR #83, opened 2016-12-19
Fix for gulp check on Windows

alecwhittington:
When running "gulp check" on a Windows machine, it would fail to run tests with issues related to Sinon and Webpack. The included module.noParse regex would only work on a Linux based machine, this commit allows Windows support


jchip:
thanks for the PR

PR #84, opened 2016-12-19
auto disable ssr based on load: Adding to Generator

animesh10:


packages/generator-electrode/generators/webapp/templates/server/conditions/machine-info.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
```
jchip: should not copy these files unless user answers yes to `Disable server side rendering based on high load?`.  Consider moving these files to a separate module.

packages/generator-electrode/generators/app/templates/_package.json
```diff
@@ -35,7 +35,8 @@
     "electrode-server": "^1.0.0",
     "electrode-static-paths": "^1.0.0",<% } else { %>
     "express": "^4.0.0",<% } %> <% if (isPWA) { %>
-    "react-notify-toast": "^0.1.3",<% } %>
+    "react-notify-toast": "^0.1.3",<% } if (isAutoSSR) {%>
+    "electrode-auto-ssr": "git+https://github.com/animesh10/electrode-auto-ssr.git", <% } %>
```
jchip: this needs to be added to https://github.com/electrode-io/electrode/blob/master/gulpfile.js#L17 for build CI test.

Also for now, change this to the version you expected to publish electrode-auto-ssr as.  ie. `^1.0.0`.  The CI test will automatically change it to point to the local copy.  Once PR is merged, we can then publish electrode-auto-ssr.

packages/generator-electrode/generators/app/index.js
```diff
@@ -169,6 +170,13 @@ module.exports = generators.Base.extend({
         },
         {
           type: "confirm",
+          name: "autoSsr",
```
jchip: the CI should test this flag of the generator. https://github.com/electrode-io/electrode/blob/master/gulpfile.js#L61-L74

packages/generator-electrode/generators/config/index.js
```diff
@@ -34,7 +39,8 @@ module.exports = generators.Base.extend({
         projectName: this.options.name,
         routeValue: routeMatch,
         pwa: this.options.pwa,
-        serverType: this.options.serverType
+        serverType: this.options.serverType,
+        autoSsr: this.options.autoSsr
```
jchip: since `isAutoSSR` is used above, should be consistent here.

packages/generator-electrode/generators/app/index.js
```diff
@@ -356,7 +365,8 @@ module.exports = generators.Base.extend({
         options: {
           name: this.props.name,
           pwa: this.props.pwa,
-          serverType: this.props.serverType
+          serverType: this.props.serverType,
+          autoSsr: this.props.autoSsr
```
jchip: since `isAutoSSR` is used above, should be consistent here.

packages/generator-electrode/generators/app/index.js
```diff
@@ -366,7 +376,8 @@ module.exports = generators.Base.extend({
     if (!this.fs.exists(this.destinationPath('server/plugins/webapp'))) {
       this.composeWith('electrode:webapp', {
         options: {
-          pwa: this.props.pwa
+          pwa: this.props.pwa,
+          autoSsr: this.props.autoSsr
```
jchip: since `isAutoSSR` is used above, should be consistent here.

packages/generator-electrode/generators/config/index.js
jchip: since `isAutoSSR` is used above, should be consistent here.

PR #85, opened 2016-12-20
add tasks to optimize react&react-dom for SS production

jchip:
Add gulp tasks to optimize `node_modules/react` and `node_modules/react-dom` into `.prod`, in preparation for removing `babel-register` in prod mode.


ananavati:
👍

PR #99, opened 2016-12-28
KoaJS support for generator-electrode

animesh10:


packages/electrode-react-webapp/package.json
```diff
@@ -52,6 +52,7 @@
     "electrode-archetype-njs-module-dev": "^1.0.1",
     "electrode-server": "^1.0.1",
     "gulp": "^3.9.1",
+    "koa-router": "^5.4.0",
```
jchip: should this be under devDependencies?

PR #107, opened 2017-01-12
TodoList sample with server side persistence.

animesh10:
Adding a todo app that is persisted on a file on the server. The contents of the app are initialized when the app is loaded, so you can continue from where you left off.
todoapp


samples/universal-react-node/server/plugins/updateStorage.js
```diff
@@ -0,0 +1,31 @@
...
+plugin.register = function (server, options, next) {
+  server.route({
+    method: "POST",
+    path: "/updateStorage",
+    handler: (request, reply) => {
+      fs.writeFile(path.join(process.cwd(), "/server/storage.json"),
+        JSON.stringify(request.payload),
+        "utf-8",
+        (err) => {
+          if (err) { throw err; }
```
jchip: need to handle this more gracefully.

samples/universal-react-node/server/views/index-view.jsx
```diff
@@ -32,11 +38,22 @@ function storeInitializer(req) {
...
       // DO ASYNC THUNK ACTIONS HERE : store.dispatch(boostrapApp())
-      Promise.resolve({})
-    ]).then(() => {
+      Promise.resolve({}),
+      new Promise((resolve, reject) => {
+        fs.readFile(path.join(process.cwd(),"/server/storage.json") , (err, data)=> {
+          if(err) {
+            reject(err);
+          } else if(data.length > 0){
+            resolve(JSON.parse(data));
```
jchip: JSON.parse could fail.  Add more promise chain steps.

samples/universal-react-node/server/views/index-view.jsx
```diff
@@ -32,11 +38,26 @@ function storeInitializer(req) {
...
+            try {
+              resolve(JSON.parse(data));
+            } catch (e) {
+              resolve({});
+            }
+          } else {
+            resolve({});
+          }
+        });
+      })
```
jchip:
```js
      new Promise((resolve, reject) => {
        fs.readFile(path.join(process.cwd(),"/server/storage.json") , (err, data)=> {
          if (err) {
            reject(err);
          } else {
            resolve(data);
          }
        })
      })
        .then(JSON.parse)
        .catch(() => {
          return {};
        })
```

samples/universal-react-node/package.json
```diff
@@ -46,6 +46,7 @@
     "electrode-archetype-react-app-dev": "^1.6.0",
     "redux-devtools": "^3.3.1",
     "redux-devtools-dock-monitor": "^1.1.1",
-    "redux-devtools-log-monitor": "^1.0.11"
+    "redux-devtools-log-monitor": "^1.0.11",
+    "superagent": "^3.3.1"
```
jchip: should use isomorphic-fetch polyfill so if browser supports it natively then it can be used.

samples/universal-react-node/client/middleware.js
```diff
@@ -0,0 +1,25 @@
+require("es6-promise").polyfill();
+require("isomorphic-fetch");
```
jchip: should use import

samples/universal-react-node/server/views/index-view.jsx
```diff
@@ -1,12 +1,14 @@
 import ReduxRouterEngine from 'electrode-redux-router-engine';
 import React from 'react';
+const fs = require('fs');
+var path = require('path');
```
jchip: use import

samples/universal-react-node/config/default.json
```diff
@@ -1,14 +1,9 @@
 {
   "plugins": {
-    "electrode-csrf-jwt": {
-      "options": {
-        "secret": "shhhhhh",
-        "expiresIn": 60
-      }
-    },
```
jchip: we need this to demo CSRF JWT.  let's figure out what's wrong and fix it.

jchip: fetch needs to pass `x-csrf-jwt` https://github.com/electrode-io/electrode/blob/master/samples/universal-react-node/client/components/csrf.jsx#L21

PR #109, opened 2017-01-12
Sync internal archetype fix

jmcriffey:
@exogen @aweary @ananavati 

Here is what we need for the internal/external sync. Before merging we need to make sure that it doesn't break the external app examples. I've already tested the internal apps with this branch. 

We can close the PR from @exogen, as this includes his changes too.


ananavati:
👍

PR #111, opened 2017-01-16
Pass userConfig to loadConfigs

cross311:
Looks like the generator is not passing through the userConfigs from the calling of the server to loadConfigs.  loadConfigs signature takes in userConfigs, but as it is right now always gets an empty config object.  The entry point for the module is passed the userConfigs, so this pull is just actually getting it to the loadConfigs method, via the Promise.resolve.


jchip:
thanks for the PR.  @animesh10 please review.

animesh10:
Looks good!

PR #117, opened 2017-01-20
Update Extract partial to use unopinionated glob

jshado1:
Based on a discussion with @ananavati in https://gitter.im/electrode-io/electrode.

~I think the [Develop Styles](http://www.electrode.io/docs/develop_styles.html) documentation should also be updated by this PR to reflect this change, but I don't see it in this repository.~ Done: electrode-io/electrode-io.github.io#377

Note: This is a non-breaking change. In sufficiently large/deep applications, this could potentially have a performance impact since it traverses all paths of the `client` directory. The impact is likely negligible though.


jshado1:
Oh, duh. There's [a link](https://github.com/electrode-io/electrode-io.github.io/blob/develop/_docs/03.1.3-develop_styles.markdown) to the markdown file in github (in a different repository).

jchip:
@ananavati is this good for merge?

PR #121, opened 2017-01-24
lerna and gulp clean

donsalari:


jchip:
```bash
rm -rf packages/*/node_modules packages/*/coverage
```

should do the trick nicely

jchip:
same thing for the samples:

```bash
rm -rf samples/*/node_modules samples/*/coverage samples/*/dist samples/*/.isomorphic-loader-config.json
```

PR #124, opened 2017-01-25
Sync internal archetypes for src/lib mode

jchip:
src/lib mode puts `client` and `server` under `src` directory and transpiled into `lib` during build step so `babel-register` is not needed anymore at run time.

@ananavati @jmcriffey @aweary


packages/electrode-archetype-react-app-dev/config/babel/babelrc-server
```diff
@@ -0,0 +1,5 @@
+{
+  "presets": [
+    "es2015-node"
```
jmcriffey: Will excluding React transforms here prevent us from using React as our template language on the back end? I know most of our examples don't do that, but I think that is a pretty common thing to do.

jchip: so far we usually put React code in client and load them from server.  for the most part, the server code don't expect to have React code.  app's `server/.babelrc` doesn't have to extend this one anyways.

PR #120, opened 2017-01-24
Allow non-GET methods on electrode-react-webapp routes

citelao:
This change lets users configure an optional `methods` attribute for `electrode-react-webapp` paths. This lets me render `POST` requests on server.

    "webapp": {
      "module": "electrode-react-webapp/lib/hapi",
      "options": {
        "pageTitle": "title",
        "paths": {
          "/{args*}": {
            "method": ["GET", "POST"],
            "content": {
              "module": "./server/views/index-view"
            }
          }
        }
      }
    }

This needs a Koa implementation, and I have not tested the Express implementation. Works for Hapi.

PR #130, opened 2017-01-30
Handle Strings passed as a method in config.

animesh10:
Fix on top of : https://github.com/electrode-io/electrode/pull/120


packages/electrode-react-webapp/lib/express/index.js
```diff
@@ -17,7 +17,10 @@ const registerRoutes = (app, options, next) => {
           registerOptions, ReactWebapp.resolveContent(v.content));
 
         /*eslint max-nested-callbacks: [0, 4]*/
-        const methods = v.methods || ["GET"];
+        let methods = v.method || ["GET"];
+        if (typeof (methods) !== "object") {
```
jchip: `Array.isArray`

Issue #119, opened 2017-01-24
Include a <noscript> in `index.htm` for graceful fail


citelao:
If a user has JS disabled, they should see an error message—not a blank screen.

animesh10:
Closing since PR is merged.

citelao:
I'm not sure this should be closed so quickly; since we have server-side rendering, a fully isomorphic webapp will only be less nice, rather than non-functional as the `<noscript>` in the pull request implies.

This might be something that users override by using their own template; maybe we can document that better?

PR #133, opened 2017-01-31
add noscript tag

animesh10:
for issue: https://github.com/electrode-io/electrode/issues/119


packages/electrode-react-webapp/lib/index.html
```diff
@@ -13,5 +13,6 @@
 <div class="js-content">{{SSR_CONTENT}}</div>
 {{WEBAPP_BODY_BUNDLES}}
 <script>if (window.webappStart) webappStart();</script>
+<noscript><p>JavaScript is Disabled</p></noscript>
```
jchip: how about something like, "sorry, this web page requires JavaScript to function properly.  Please enable JavaScript in your browser and reload".

PR #136, opened 2017-01-31
remove extractTextPlugin in hot mode

### LINKED ISSUE #135: [HMR] css are not updated in hot mode
related to #125

orlandodemauro:
fixes #135 

@jchip please review

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract.js
@@ -61,7 +61,7 @@ module.exports = function () {
       loaders.push({
         name: "extract-stylus",
         test: /\.styl$/,
-        loader: hmr ? "style!css??-autoprefixer!stylus-relative-loader" : ExtractTextPlugin.extract(styleLoader, stylusQuery, {publicPath: "" })
+        loader: hmr ? "style!css?-autoprefixer!stylus-relative-loader" : ExtractTextPlugin.extract(styleLoader, stylusQuery, {publicPath: "" })
       });
     }
 
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract.js
@@ -49,7 +49,7 @@ module.exports = function () {
   return function (config) {
     var stylusQuery = cssLoader + "?-autoprefixer!" + stylusLoader;
     var cssQuery = cssLoader + "?modules&-autoprefixer!" + postcssLoader;
-
+    var hmr = process.env.HMR === "true";
     // By default, this archetype assumes you are using CSS-Modules + CSS-Next
     var loaders = [{
         name: "extract-css",
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract.js
@@ -54,14 +54,14 @@ module.exports = function () {
     var loaders = [{
         name: "extract-css",
         test: /\.css$/,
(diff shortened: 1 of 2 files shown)


orlandodemauro:
@jchip I don't think that the failure is related to my changes, can you confirm it please?

jchip:
does this mean you can't use CssModule and HMR together?

orlandodemauro:
yes

orlandodemauro:
@jchip I also noticed that 

```
  if (module.hot) {
    module.hot.accept();
  }
```

has not been added into the generator or app in #128.

jchip:
hi, thanks for looking into this.

I found that the `module.hot.accpet()` is not really needed.

Please let me think about this since HMR and css module breaks.  I am not sure if it's a good idea to just break CSS module for HMR.

orlandodemauro:
@jchip thanks for.your feedback. What do you mean with CSS module breaks? It working for me

jmcriffey:
@orlandodemauro See my comment here: https://github.com/electrode-io/electrode/issues/135

I think your issue can be solved by using `electrode-archetype-react-app/supports` and the `cssModuleHook` in your server setup.

jmcriffey:
Re-reading your issue and fix, my previous comments may be off base. I'm not sure why you would have issues with HMR and ExtractTextPlugin though. I suggest though that we dig deeper and figure out why it isn't working before removing it though.

orlandodemauro:
@jchip @jmcriffey 
Maybe I'm misunderstanding something but here my thoughts. We have two issues:

1. the css are not hot reloaded

This issue is related to the use of the ``` extract-text-webpack-plugin ```. Hot reloader is not working with extractTextPlugin and it shouldn't  be used in dev as in comment by the author of webpack https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/30#issuecomment-107242576

about css module break, I don't see why it should be broken. CSS Module can be used without the extract-text-webpack-plugin.

2. ``` cssModuleHook ``` is not working properly even using 
```
supports.cssModuleHook({
  generateScopedName: '[name]__[local]___[hash:base64:5]'
});

```

the classes are rendered as ``` _1cnXHAlzCtH79wfQ7FYQFz ``` and not as ``` header__container_Z56Gh ```. Checking the demo config https://github.com/css-modules/css-modules-require-hook/tree/master/demo I found this note:

> to generate the same names in runtime as the CSS styles you should provide the same pattern to webpack and to the require hook.

for this on my branch https://github.com/orlandodemauro/electrode/tree/fix_css_modules_require_hook I set up ``` localIdentName ``` in webpack. and now the cssHook is working properly for me.

I have both fixes locally and everything is working as expected.

I hope this can help

orlandodemauro:
This PR fixes only issue 1

jchip:
@orlandodemauro thanks for the detail info.  @jmcriffey and @ananavati may have more insights regarding the issues.

jmcriffey:
@orlandodemauro My comments before are directed at solving the 2nd issue. I like your solution with adding in a config for the scope name, but it still has an issue as you have it in your branch.  In your fix, you are not setting the rootDir on the server. This is an absolutely crucial part, as your server and client will generate different hashes based on that rootDir and you will end up with a flash of un-styled content if your scoped name contains part of the hash. 

As for the first issue, I understand that removing the extract plugin fixes your issue, but I am unsure as to why that is the case. I would prefer that we find a less destructive solution to the problem. If we investigate further and find that removing the extract plugin is the only solution, then I think we need to do a bit of restructuring so that this code is only run in the `webpack.config.hot.js` config file instead of relying on an environment variable.

orlandodemauro:
@jmcriffey @jchip hi guys, any news about css hot reload? thank you

karlhorky:
@orlandodemauro there are some conflicts in this pull request now, can you fix these?

PR #140, opened 2017-02-01
Stylus and css modules together

jmcriffey:
@jchip @ananavati @aweary 

This PR adds the ability for an app to use both stylus and css modules together. While adding in this feature, I found that there is no reliable way to detect that both css modules and stylus should be used (since a project could contain only stylus files but still want to use css modules), so I have added in reading from a `config/css-config` file. I believe we can also utilize the same config file to allow options like a custom `generatedScopeName`.

### DIFF
--- a/packages/electrode-archetype-react-app/config/archetype.js
+++ b/packages/electrode-archetype-react-app/config/archetype.js
@@ -137,20 +137,20 @@ function loadDev() {
     devDir,
     devPkg,
     devRequire,
-    webpack: Object.assign({}, archetypeOptions.webpack, {
+    webpack: Object.assign({}, {
       devHostname: "localhost",
       devPort: getInt(process.env.WEBPACK_DEV_PORT, 2992),
       testPort: getInt(process.env.WEBPACK_TEST_PORT, 3001),
       modulesDirectories: []
-    }),
-    config: Object.assign({}, archetypeOptions.configPaths, {
+    }, archetypeOptions.webpack),
+    config: Object.assign({}, {
       babel: `${configDir}/babel`,
       eslint: `${configDir}/eslint`,
       istanbul: `${configDir}/istanbul`,
       karma: `${configDir}/karma`,
       mocha: `${configDir}/mocha`,
       webpack: `${configDir}/webpack`
-    })
+    }, archetypeOptions.configPaths)
   });
 }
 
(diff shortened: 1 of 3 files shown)


jchip:
I've been thinking of adding a documented approach to allow setting configs for the archetype.  there are a few options

  - leaving config file in the `config` like you did, but that sits next to the app's config.  
  - add a new directory `archetype` and loads a `config.js` from there that serves as the single location for configuring.  in addition, we can use that directory to host any other customization we support for the archetypes.

There's always the `process.env` settings and user an set that in their `gulpfile.js`.

Any other input over those options?

jmcriffey:
I like the idea of having a single config file from your second option. I think that approach is clearer for the end user and much easier to document. 

As far as environment variables go, I think most of the config settings that we want to set belong elsewhere. If we use them in one or two places now, that is fine, but there is a scenario in which we end up with too many options to sanely keep in the environment.

jchip:
could you update this PR to load from `PWD/archtype/config`?  We should just merge that into our `config/archetype.js`.  So user can override settings in `config/archetype.js`.

jmcriffey:
Yep, PR update coming in shortly.

jmcriffey:
@jchip This is ready for re-review. I am reading from `CWD/archetype/archetype.js` and adding that to `archetype.options`. Let me know if that could be named more descriptively.

packages/electrode-archetype-react-app/config/archetype.js
```diff
@@ -121,20 +137,20 @@ function loadDev() {
...
     devRequire,
-    webpack: {
+    webpack: Object.assign({}, archetypeOptions.webpack, {
       devHostname: "localhost",
       devPort: getInt(process.env.WEBPACK_DEV_PORT, 2992),
       testPort: getInt(process.env.WEBPACK_TEST_PORT, 3001),
       modulesDirectories: []
-    },
-    config: {
+    }),
```
aweary: Do we want `devHostname` and the other options here to be configurable? Because the current order would cause these keys to always be overwritten. I think we want:

```js
Object.assign({
  devHostname: "localhost",
 ...
}, archetypeOptions.webpack)
```

jmcriffey: Yep that is a good catch. I definitely meant for those to be reversed.

Issue #145, opened 2017-02-04
🐞 try/catches swallow application errors during build


jshado1:
There are a bunch of `try` `catch` blocks around `require()`s in the Electrode source that cause application errors to fail silently, hiding the invaluable errors that the underlying build-tools throw. These should be replaced with fs checks.

jchip:
all of them are for checking optional files.  Using fs checks would have the same behavior.  Is there a specific scenario you can offer?  But thanks for raising the issue, we do need to improve logging to let user know what the build is doing with regards to this.

jshado1:
An fs would not have the same behaviour:
```js
if (fs.fileExists(optionalFile)) {
  require(optionalFile); // optionalFile contains code that throws an error --> error thrown and execution stops as expected
} else someLogger.debug(`${optionalFile} not found`);
```
Vs
```js
try {
  require(optionalFile);
} catch (err) {}
```
Error is swallowed.
Vs
```js
try {
  require(optionalFile);
} catch (err) {
  someLogger.debug(err); // error is not swallowed, but exection does not stop
}
```

The code that includes optional webpack.config files uses a try/catch, which swallows errors thrown in them. That was a fun waste of an afternoon tracking down.

Now I have a React module that Electrode reports can't be found; yet strangely when I log `process.cwd()` and take the outputted path of the supposedly non-existant file, it does in fact exist: `cat "$1$2"`. So I suspect the module is actually throwing an error that Electrode is swallowing with another try/catch and instead reports Module Not Found.

jshado1:
Safeguarding against an application error crashing the server is desirable for prod, but not in dev. In dev, I expect it to explode.

jchip:
ah, yeah, thanks for pointing that out.

PR #148, opened 2017-02-04
archetype-react-app: more informative optional require

jchip:
resolves https://github.com/electrode-io/electrode/issues/145

@jshado1


jshado1:
I would suggest something like [Winston](https://www.npmjs.com/package/winston) for logging instead just `console` because it supports logging levels, allowing one to dis/enable various types of logging globally (say, per environment, or with an environment variable). It would also allow "users" of Electrode to remotely store the logging data.

Maybe for V2?

packages/electrode-archetype-react-app-dev/config/webpack/add-dll-references.js
```diff
@@ -25,7 +25,7 @@ module.exports = function (config) {
       }));
     }
   } catch (err) {
-    //eslint-disable-line no-empty
+    console.log("add-dll-references failed", err);
```
jshado1: `console.error` ? this won't abort execution, but it will output like a fatal error (and get colourised red)

jchip: I assumed that this was not a fatal error so the log is just FYI to the user.

PR #155, opened 2017-02-08
Refactor Webpack config to allow apps to override config partials

Nickersona:
## Context
Our team was struggling with moving over a project using SASS  to work within the confines of CSS modules and the existing configuration options (via webpack.config.*.js files in our project root) wouldn't allow us to replace the css loader, causing all kinds of hacks and work arounds, to make SASS viable. Writing our own archetype seems to be the given solution, but I'd rather not have to maintain my own archetype when I otherwise agree with the rest of the conventions provided `electrode-archetype-react-app`.

### TL;DR; Define your own, project specific Webpack partials, and have them replace what's defined in the archetype.

I have a modification to the base webpack config file that allows projects using electrode to specify webpack partials in the projects config directory(`config/webpack/partial/`) with the same names as partials in  `node-modules/electrode-archetype-react-app-dev/config/webpack/partials` which replaces what's in the archetype.

I've added a sequence property to each of the partials that were `required()` by `config/webpack/base.js` so that when it globs the ./partial directory it only grabs those partials with the correct sequence, (in this case 0 since it's base config object). This is a pretty MVP feature, but I  figure that sequence could be used down the line to define which partials are required at which stage of the webpack config construction process.  Overriding files in a projects config directory, would also need this sequence property.

This will require some documentation which I'm happy to write, but I wanted to get the code in front some of the core contributors to see if this is something worth developing.  I've tried to touch the partial files as little as possible so they continue to work as expected with the rest of the webpack config files. Tests are passing 🎉

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/webpack/base.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/base.js
@@ -3,26 +3,21 @@
 var _ = require("lodash");
 var fs = require("fs");
 var mergeWebpackConfig = require("webpack-partial").default;
+var glob = require("glob");
 
 // config partials
-var babelConfig = require("./partial/babel");
-var extractStylesConfig = require("./partial/extract");
-var fontsConfig = require("./partial/fonts");
-var imagesConfig = require("./partial/images");
-var statsConfig = require("./partial/stats");
-var isomorphicConfig = require("./partial/isomorphic");
-var jsonConfig = require("./partial/json");
-var pwaConfig = require("./partial/pwa");
 var archetype = require("../archetype");
 var Path = archetype.Path;
 var AppMode = archetype.AppMode;
 var context = Path.resolve(AppMode.src.client);
+var configContext = Path.resolve(AppMode.src.dir);
 
 var archetypeNodeModules = Path.join(archetype.dir, "node_modules");
 var archetypeDevNodeModules = Path.join(archetype.devDir, "node_modules");
 var inspectpack = process.env.INSPECTPACK_DEBUG === "true";
 
 var optionalRequire = require("optional-require")(require);
+
 
 /* eslint-disable func-style */
 
@@ -83,14 +78,41 @@ var baseConfig = {
   }
 };
 
-module.exports = _.flow(
+
+const basePartials = {};
+glob.sync(Path.join(__dirname, 'partial/*.js')).forEach(function(filePath) {
+  const fileName = Path.parse(filePath).name;
+  const partial = require(filePath);
+  // Potentially sort order of partials by sequence in the future. If multiple partials of equal
+  // sequence, assume order is not important.
+  if(partial.sequence === 0) {
+    basePartials[fileName] = require(filePath);
+  }
+});
+
+const customPartials = {};
+glob.sync(Path.join(configContext, 'config/webpack/partial/*.js')).forEach(function(filePath) {
+  const fileName = Path.parse(filePath).name;
+  const partial = require(filePath);
+  if(partial.sequence === 0) {
+    customPartials[fileName] = require(filePath);
+  }
+});
+
+const mergedPartials = _.merge(basePartials, customPartials);
+const partials = _.reduce(
+    mergedPartials,
+    function(results, func) { 
+      results.push(func()); 
+      return results;
+    },
+    []
+);
+
+var configPartials = _.concat(
+  [],
   mergeWebpackConfig.bind(null, {}, baseConfig),
-  babelConfig(),
-  extractStylesConfig(),
-  fontsConfig(),
-  imagesConfig(),
-  statsConfig(),
-  isomorphicConfig(),
-  jsonConfig(),
-  pwaConfig()
-)();
+  partials
+);
+
+module.exports = _.flow(configPartials)();
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js
@@ -6,7 +6,9 @@ var AppMode = archetype.AppMode;
 var Path = archetype.Path;
 var _ = require("lodash");
 
-module.exports = function (babel) {
+
+
+module.exports = Object.defineProperties(function (babel) {
   // regex \b for word boundaries
   var babelExcludeRegex = new RegExp(`(node_modules|\\b${Path.join(AppMode.src.client, "vendor")}\\b)`);
   return function (config) {
@@ -35,4 +37,8 @@ module.exports = function (babel) {
       }
     });
   };
-};
+}, { 
+  sequence: {
+    value: 0
+  }
+});
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract.js
@@ -41,7 +41,7 @@ if (stylusExists && !cssNextExists) {
   cssModuleSupport = false;
 }
 
-module.exports = function () {
+module.exports = Object.defineProperties(function () {
   return function (config) {
     var cssModuleStylusSupport = archetype.webpack.cssModuleStylusSupport;
     var stylusQuery = cssLoader + "?-autoprefixer!" + stylusLoader;
@@ -99,4 +99,8 @@ module.exports = function () {
       ]
     });
   };
-};
+}, {
+  sequence: {
+    value: 0
+  }
+});
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/fonts.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/fonts.js
@@ -5,7 +5,7 @@ var mergeWebpackConfig = require("webpack-partial").default;
 var urlLoader = require.resolve("url-loader");
 var fileLoader = require.resolve("file-loader");
 
-module.exports = function () {
+module.exports = Object.defineProperties(function () {
   return function (config) {
     return mergeWebpackConfig(config, {
       module: {
@@ -24,4 +24,8 @@ module.exports = function () {
       }
     });
   };
-};
+}, {
+  sequence: {
+    value: 0
+  }
+});
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/images.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/images.js
@@ -13,7 +13,7 @@ function getCdnLoader() {
   return loader && require.resolve(loader) || "file-loader";
 }
 
-module.exports = function() {
+module.exports = Object.defineProperties(function() {
   return function(config) {
     return mergeWebpackConfig(config, {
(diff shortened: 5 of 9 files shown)


jshado1:
This seems to also make it easier (possible?) to add new Archetypes: Just drop them into that `<app>/config/webpack/partial`.

By the way, is the directory name in the app archetype package supposed to be `partial` singular instead of `partials`?

jchip:
thanks for the PR.  this is the direction we are thinking also. 

For example, in preparation for further user config and customization, there was another merged PR that added some code to make `CWD/archetype` directory as a place for users to add custom configs.  

Could you please update to align with that?

Sorry I've been busy with our internal stuff, but once you update that, I should be able to look at your PR further over the weekend.

Thanks.

jchip:
what I was thinking for user webpack partials configs.

`CWD/archetype/config/webpack/<mode>/*.js`

where `<mode>` can be `base`, `dev`, `coverage`, `hot`, etc, corresponding to our webpack configs.

is that how your PR works?

jshado1:
@jchip, yes: our PR globs Electrode's archetypes and the app's, and creates a collision via `_.merge(electrodePartials, appPartials)`, whereby Electrode's are overrided by the app's:

```
├─┬ electrode-archetype-react-app-dev/config/webpack/partial/
  ├── babel.js
  ├── …
  ├── extract.js
  └── …
```
```
├─┬ <app>/config/webpack/partial/
  ├── some-new-archetype.js
  └── extract.js
```

`<app>/config/webpack/partial/extract.js` overrides `electrode-archetype-react-app-dev/config/webpack/partial/extract.js` and `<app>/config/webpack/partial/some-new-archetype.js` is added (assuming there is no collision). Note that only `webpack/partial` is globbed (applies to both Electrode and App).

The reason we went with `<app>/config/webpack/partial/` is to match Electrode's existing directory structure's convention.

Why would you throw in an extra `<app>/archetype/` when `<app>/config/` already exists? Also what would go in `<app>/archetype/` besides `config/`?

jshado1:
Correction: No. This affects only `base` and the partials it originally imported (those partials now have a `sequence` property to identify them).

jchip:
`<app>/config` is generally for the Node/Hapi server/app configuration used for run time.

jchip:
we are planning to add more customization for the archetype and it's more manageable to keep that separated.

jmcriffey:
I have recently tried to rectify this issue in a different way by extending loader configurations. For example, I needed to exclude a different set of folders for the babel loader. It seems like a pretty simple solution to allow each loader to be extended. Would being able to extend a loader config have solved the issue you had? Keep in mind you can already extend parts of the config (plugins, output, etc) via a base level webpack config. It is just not currently possible to extend loaders. 

I think you will get into trouble quickly trying to override entire partials. Some partials are simple, but others are quite complicated. I feel like you are more likely going to want to change a few specific parts of a configuration rather than replace a whole partial. 

If you wanted to make a small change using this method, you'd pretty much need to copy and paste the existing partial and make edits, correct? That would mean you would miss out on bug fixes and new features. Does that make sense?

jshado1:
@jchip if `<app>/config` is just for server, why not put it under `<app>/[src/]server`?

@jmcriffey `webpack.config.js` does not always get used. We've currently hacked an extend together for every possible `webpack.config[.mode].js`, but that's a pain and fragile:
```js
// webpack.config.js, webpack.config.coverage.js, webpack.config.dev.js, etc…

// …
const baseWebPack = require('./webpack.config.base') || {};

module.exports = _.merge(baseWebPack, {

});
```

It sounds like your solution might work for some, but I'd have to see it. For this specific case, we needed to remove the postCSS/CSS Modules partial (`extract.js`) altogether.

jchip:
For sure we don't want to mix server config and archetype config.  The structure of the `<app>` at the root level is to support server in production.  The archetype is intended for build/dev time.  `<app>/src` is transpiled into `<app>/lib`.   We started with `<app>/config` for server and we want to keep priority for the server when it's running in production.  No need to mix that up because we want to add more customization for archetype.

citelao:
I want to jump in on this since CSS Modules are causing me headaches.

@jshado1 @jmcriffey If the problem is overriding loaders, can we focus on that rather than overriding partials? It seems to me that there are a couple larger problems:

1. Cannot globally override webpack config; must do it for `.dev`, `.hot`, etc
2. Cannot extend sections of webpack config (`loaders`, `plugins`, etc) easily
3. Cannot override arbitrary sections (cannot update the settings of `extract-css` loader, e.g.

The simplest solution to these problems IMO:

1. Allow override of `base.js` the same way the other webpack configs are handled.
2. Bring `mergeWebpackConfig` to the frontend
3. Extend (?) `mergeWebpackConfig` to check the `name` property of new loaders and plugins

Basically, why go through the hoopla of overriding partials? That seems more of a code-org decision than a true structural one. I'd rather override `base.js`.

citelao:
IRT `config/` vs `archetype/`, current `webpac.config.*.js` overrides live in the project root---should they be moved to `archetype/` as well? 

I agree with @jchip that `config/` is not an appropriate location since it refers to server/app configuration rather than build configuration. The name `archetype/` is growing on me as a choice, too.

jshado1:
@jchip @citelao What I meant was that `<app>/config/` might be moved into `<app>/server/` if it only applies to server, and then `<app>/archetype/` would live wherever. Often there's a `<app>/build/` directory that contains all the scripts etc related to building the application. So you'd have something like:

```
app
├─┬ build
  ├─┬ webpack
    └─┬ partials
      └── extract.js
  ├── webpack.config.js
  └── webpack.config.dev.js
├── …
├─┬ server
  ├─┬ config
    ├── default.js
    ├── development.js
    └── production.js
  ├── express-server.js
  └── index.js
├── …
└── package.json
```

citelao:
@jshado1 I'm not a fan of `build/` since a lot of projects use that to hold temporary artifacts. I think what @jchip was saying (and what I agree with) is that `config/` stores *app* configuration, so it makes sense to live at the top (since the app is the entire webapp).

Maybe (since I think partials override is a bad thing):

```
app
├─┬ archetype
  ├─┬ eslint
    └── ...
  ├── base.js
  ├── webpack.config.js
  └── webpack.config.dev.js
├┬ config
 ├── default.js
 ├── development.js
 └── production.js
├── …
└── package.json
```

Nickersona:
I'm having a hard time seeing why `/archetype` wouldn't live under `/config`. Something like?

```
app
├┬ config
  ├┬ archetype
   ├─┬ eslint
       └── ...
   ├── base.js
   ├── webpack.config.js
   └── webpack.config.dev.js
  ├┬ server
   ├── default.js
   ├── development.js
   └── production.js
├── …
└── package.json
```

jshado1:
@citelao there's already `tmp`, which is where temporarily artefacts should go 😜 

`config/default.js` contains things like port numbers and cors dis/enabling that are used by the express server. That sounds pretty server-centric to me.

citelao:
@jshado1 I think my point is that the entire app is basically the server (since it's universal), so configuration should be at root. Besides, the whole `server/` dir is moving to `src/server/` and that's definitely not a nice place for config IMO. 

Still don't like `build/` because at first glance it's still confusing, though I see your `tmp` point.

I'd much prefer @Nickersona 's suggestion of having them live in `config/archetypes/`, especially if we move towards loader-overriding rather than partial-overriding

jshado1:
I'm fine with @Nickersona's suggestion, but at that point, why not just mirror the src directory and call it `config/client/` and `config/server/`?
```
app
├─┬ config
  ├─┬ client
    ├─┬ partials // this name is very confusing
      └── extract.js
    ├─┬ eslint
      └── …
    ├── base.js
    ├── webpack.config.js
    └── webpack.config.dev.js
  └─┬ server
    ├── default.js
    ├── development.js
    └── production.js
├─┬ src
  ├─┬ client
    └── …
  └─┬ server
    └── …
├── …
└── package.json
```

Why are they called "partials" by the way? They seem to be processors or loaders.

citelao:
@jshado1 `config/client/` and `config/server/` I don't like because, although webpack builds the client-side, it's also used for the server. 

It also makes the whole app look like `folder/client` and `folder/server`---maybe there's a structural issue there?

 But it brings up a good point---confippet isn't available on the client. It doesn't belong in a "global" `config/` directory.

I think they're called partials because they deal with a fixed segment of configuration (css loading, hot reloading, etc), but I reiterate that overriding seems to be done on a per-loader or -plugin basis, not on a partial basis.

I just discovered `src/` myself 😁

jshado1:
@citelao is webpack actually used for the server? We're actually trying to solve a different (but possibly related) problem right now that appears to be caused by the opposite—that it is not run thru webpack.

citelao:
@jshado1 I thought the `lib/` build passed through, but I haven't looked into that build mode TBH. But calling webpack "client config" and confippet "server config" just feels wrong since they're *so* orthogonal

jchip:
we are switching app archetype to webpack 2 and just published a major 2.0 version.  We can continue this on the 1.0 branch or move it over to the new 2.0 branch.

Regarding the directory config, I prefer to just leave that alone and reserve another location for archetype customization.

jchip:
@citelao the webpack 2 switch is mostly transparent, except a few quirks that's specific with webpack 2.  Our samples work by just switching to the new webpack 2 archetypes.  We are planning on tackling code splitting and we don't want to do that on webpack 1.

jchip:
Thinking about this some more.  One potential issue (correct me if I am mistaken) is that the partials order are now dependent on the file names or how glob return them?

jshado1:
@jchip I think that is not entirely correct because you can use `sequence` to specify an explicit order; however, the order of partials doesn't seem to (and I think _shouldn't_) matter. But without forcing an order with `sequence`, yes, the order would be determined by the glob.

citelao:
@jchen that's why I think it's better to focus on confit-based overloads. The partials strike me as an organizational thing, and, as was mentioned elsewhere in the thread, overriding them means losing all the bug fixes etc they might have. 

Webpack 2's layout is much more conducive to loader-specific override (esp ExtractText). Could we not build on the webpack merge features you already have?

That gives two options: if you like the archetype but want to change some options, you can do it without losing ALL the partial, and if you don't like the archetype you can continue writing your own

jshado1:
@citelao ideally, both extending and overriding would be possible. In our case, we're needing the override because one of the partials causes adverse behaviour and we want to remove it altogether.

citelao:
@jshado1 ok, that changes my tune. I thought the partials didn't/shouldn't have side effects. 

Both!

citelao:
@cross311 As you mentioned in #205, the larger `archetype/` question is "how do we override any arbitrary part of the archetype?"

Does that discussion belong here?

jchip:
creating a flexible webpack config composer to address this https://github.com/electrode-io/electrode/pull/222

jchip:
rewritten all webpack partials https://github.com/electrode-io/electrode/pull/226 with the new webpack-config-composer.  This will allow you to enable/configure/override any of the archetype's partials, or add your own.

PR #205, opened 2017-02-22
Single Quotes Question

cross311:
I would like the ability to use single quotes with electrode generated code.  Adding a question around if use wants to do this as well.  Then running the fix command from eslint to switch the double quotes to single quotes.

### DIFF
--- a/packages/generator-electrode-component/app/index.js
+++ b/packages/generator-electrode-component/app/index.js
@@ -34,10 +34,10 @@ var ReactComponentGenerator = yeoman.generators.Base.extend({
 			_.extend(this, props);
 			this.packageName = _.kebabCase(_.deburr(this.projectName));
 			this.componentName = this.packageName
-        .replace(/^\s+|\s+$/g, "")
-        .replace(/(^|[-_ ])+(.)/g, function (match, first, second) {
-          return second.toUpperCase();
-        });
+			.replace(/^\s+|\s+$/g, "")
+			.replace(/(^|[-_ ])+(.)/g, function (match, first, second) {
+			return second.toUpperCase();
+		});
 			this.currentYear = new Date().getFullYear();
 			done();
 		}.bind(this));
@@ -152,7 +152,7 @@ var ReactComponentGenerator = yeoman.generators.Base.extend({
 		this.log(
 			"\n" + chalk.green.underline("Your new Electrode component is ready!") +
 			"\n" +
-      "\nYour component is in src/ and your demo files are in demo/" +
+			"\nYour component is in src/ and your demo files are in demo/" +
 			"\n" +
 			"\nType " + chdir + "'gulp demo' to run the development build and demo tasks." +
 			"\n"
--- a/packages/generator-electrode-component/app/index.js
+++ b/packages/generator-electrode-component/app/index.js
@@ -140,7 +140,7 @@ var ReactComponentGenerator = yeoman.generators.Base.extend({
 			this.template("src/_Component.js", "src/index.js");
 		},
 		test: function() {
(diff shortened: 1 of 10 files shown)


packages/generator-electrode-component/app/templates/_gulpfile.js
```diff
@@ -18,6 +18,9 @@ const tasks = {
   "generate": ["generate-metadata", "generate-documentation"],
   "generate-documentation": () => exec(`electrode-docgen --package ./package.json --src ./src --markdown components.md`),
   "generate-metadata": () => exec(`electrode-docgen --package ./package.json --src ./src --metadata components.json`),
+<% if (quoteType === "'") { %>
+  'archetype:lint-server': () => exec(`eslint --color -c ./.eslintrc config/karma config/webpack demo-server`),
```
cross311: this was the only place i saw to hook in.  But seems a little sketch since it is overwriting an archetype task.

packages/generator-electrode-component/app/templates/demo/_demo.jsx
```diff
@@ -1,5 +1,4 @@
 /*@flow*/
-/*global document:false*/
```
cross311: removed this because it was not in use and it was failing the eslint.

jchip:
I understand.  When I started at wml my muscle memory was single quote.

I think the right approach is to make it easier to have your own eslint configs.

Of course the generator still generates code from templates that have double quotes.  I think for that I might use something like prettier to reformat the code after it's generated.

packages/generator-electrode-component/app/index.js
```diff
@@ -136,6 +145,9 @@ var ReactComponentGenerator = yeoman.generators.Base.extend({
 	},
 
 	end: function() {
+		if (this.quoteType === "'") {
+			this.spawnCommandSync("node_modules/.bin/eslint", ["--fix", "src", "demo", "example", "test", "--ext", ".js,.jsx"]);
```
jchip: right, I forgot that eslint can do this.  so no need to use prettier.

Is eslint added to the `dependencies` though?

packages/generator-electrode-component/app/templates/_gulpfile.js
jchip: overwriting the task is ok, but I don't remember why these tasks are not in the component archetype.

packages/generator-electrode-component/app/index.js
cross311: yes it is in the node_modules because of the archtypes. so you have to run this step after the this.npmInstall().

cross311: Only kicker is that if there is a eslint issue with the generated code it will not remove the double quotes.

packages/generator-electrode-component/app/templates/_gulpfile.js
cross311: actually i might have to fix a couple more of these hooks. that just does the demo server.

cross311:
@jchip Yes to easier to extend the eslint config.  I know your team is looking into and `archtype` folder for webpack.configs and things.  is there a way to leverage that with eslint?

packages/generator-electrode-component/app/templates/_gulpfile.js
jchip: OK, do you want me to hold off merging the PR for now?

citelao:
@cross311 I think the `archetype/` folder would allow you to override any part of the archetype. A plain `eslint` file wouldn't be good enough, though. As @jchip said, the generated code would have to be `--fix`ed.

cross311:
@jchip Alright I have covered all the other eslint entry points in the gulpfile

Issue #116, opened 2017-01-20
Greedy string replace in copy-as-flow script


bradbarrow:
If you generate a component in a directory that has a directory called `src` in it's parent path, the script will fail.

e.g. 

```
$ cd /Users/brad/src/apps/something/components
$ yo electrode-component
```
`copy-as-flow` will fail above because it tries to copy from `/Users/brad/*lib*/apps/something/src/something`

https://github.com/electrode-io/electrode/blob/master/packages/electrode-archetype-react-component/scripts/copy-as-flow-declaration.js#L30

PR #209, opened 2017-02-24
fix issue #116: Greedy string replace in copy-as-flow script

ananavati:
fixes https://github.com/electrode-io/electrode/issues/116


jchip:
👍

PR #211, opened 2017-02-24
Support stylus modules for component archetype

aweary:
Similar to what we did in https://github.com/electrode-io/electrode/pull/140 for app archetype.

cc @jmcriffey


packages/electrode-archetype-react-component/config/webpack/partial/styles.js
```diff
@@ -10,6 +10,11 @@ const cssnext = archDevRequire("postcss-cssnext");
 const styleLoader = archDevRequire.resolve("style-loader");
 const cssLoader = archDevRequire.resolve("css-loader");
 const postcssLoader = archDevRequire.resolve("postcss-loader");
+const stylusLoader = archDevRequire.resolve("stylus-relative-loader");
+
+const configPath = Path.resolve("config.json");
```
aweary: We don't currently have any configuration for components like we do apps, AFAICT. `config.json` is probably not what we want to do, but I want to get some feedback before implementing anything more complicated.

jchip: The direction we are going is to have an `archetype` directory that lets user put any archetype related config/customization there.

aweary:
@jchip I updated the Webpack partial to load the config from `archetype/config.js` and updated the yeoman generator to include that file by default.

jmcriffey:
LGTM. We are going to need this in the 1.0 branch. How is that usually handled @jchip?

jchip:
Please PR this into 1.0 branch if you need it.

PR #215, opened 2017-02-27
Stylus modules component for 1.0 branch

aweary:
Same as https://github.com/electrode-io/electrode/pull/211 but for the `webpack-1.0` branch


aweary:
@jchip we need these changes for some internal stuff, so let me know what else needs to be done for this to get released soon.

PR #224, opened 2017-03-06
fix production server

SampsonCrowley:
Production server doesn't build.

fixed where it looks to be the gulp build output.

this will now work on heroku


SampsonCrowley:
How Is my thing failing? I literally only modified the scripts in package.json to actually look in the correct place

jchip:
thanks for the PR

SampsonCrowley:
Glad I could help

Issue #214, opened 2017-02-27
Electrode Component Generator doesn't allow different github repo and npm package name - [Regression]


workingBen:
The currently published and out of date electrode documents show a question in the electrode component generator that was removed in version `1.2.4`. `What is the name of the GitHub repo this will be published at?`

When we had that question, I could have a repo located at `react/tempo-modules` and a package name `@walmart/wmreact-tempo-modules` because there were two questions, one for each value. Now there is only one question, `What is your npm package name?` which is used for both values. This means we cannot create components with this naming pattern. 

SCREENSHOT OF ISSUE:
<img width="1284" alt="github-repo-collapsed-with-npm-package-name" src="https://cloud.githubusercontent.com/assets/333554/23371998/be6dcf64-fcce-11e6-81fa-42ce7566224f.png">

SCREENSHOT OF DIFF:
<img width="637" alt="screen shot 2017-02-27 at 9 11 06 am" src="https://cloud.githubusercontent.com/assets/333554/23372009/c4424050-fcce-11e6-9399-8a0e840c6ba5.png">

Please advise if I should submit a PR for a fix, or if another plan is better. Thanks!

animesh10:
Hi, the component generator has a question `What will be the GitHub organization username (e.g., 'walmartlabs')?` this is your github Org name.
The repo Url is a combination of the above org name and the package name you provide.

workingBen:
@animesh10 I want to generate a component with these settings: 

github repo name: "grocery-products-page"
github org name: "OneWalmart"
package name: "@walmart/wmreact-grocery-products-page"

Can you help me achieve this?

animesh10:
So your package URL would be `"url": "https://github.com/OneWalmart/grocery-products-page.git"` ?
But the package `name` would be `@walmart/wmreact-grocery-products-page` ?

At this time we are using the package name as the github repo name. So answering `grocery-products-page` for the `package name` and `WalmartOne` for the `OrgName` will give you `https://github.com/OneWalmart/grocery-products-page.git`, but you will have to manually edit the `name` in the `package.json`.

workingBen:
@animesh10 I have been manually editing the package name in `package.json` as you mention. Since this is a common pattern, and was supported by a previous version of the electrode component generator, I was hoping it could be added back in.

Here's an example: https://gecgithub01.walmart.com/react/tempo-modules/blob/master/package.json
SCREENSHOT: 
[...]

PR #225, opened 2017-03-06
Accept different repo and package name for component generator

animesh10:
@ananavati @workingBen
With reference to issue214: https://github.com/electrode-io/electrode/issues/214

Before: 
screen shot 2017-03-06 at 3 48 23 pm

After: 
screen shot 2017-03-06 at 3 37 01 pm


workingBen:
LGTM.

My only comments would be about preferring ES6 > ES5, but it seems this whole generator is in this style so I'm punting on all those comments. Nice work!

One nice-to-have would be **BEFORE** and **AFTER** screenshots in the PR to show the difference from a user perspective.

jchip:
merging.  In the future, when doing auto code reformatting, please be sure to separate the pure reformats in a commit by itself so it's easier to see what are the actual changes.

PR #226, opened 2017-03-06
composable webpack partials

jchip:
A sample of what user can do:

webpack.config.js:

```js
module.exports = function(composer, options) {
  const customProfile = {
    partials: {
      "_dll": { enable: false }, // turns off DLL partial
      "_images": { options: { cdnLoader: "custom-cdn-file-loader" } }, // override images file loader
      "custom-entry-partial": {
         order: 201 // right after archetype's _entry partial 
      }
    }
  };

  composer.addPartials( {
    "custom-entry-partial": {
      config: {
        main: "./my-app.js"
      }
    },
    "_output": {
      config: {
        output: {
        }
      },
      addOptions: {
        method: "replace" // replace archetype's _output partial
      }
    }
  });

  return composer.compose({}, [].concat(options.profileNames).concat([customProfile]); 
};
```


animesh10:
👍

Issue #227, opened 2017-03-07
electrode-archetype-react-component* dependencies outdated


DavidOrchard:
The dependencies are very outdated, such as "eslint": "^1.7.0" whereas app is more update to date "eslint": "^3.9.1". Among other things, this causes incompatibilities in custom eslint rules for a component that is then included in an app.

cross311:
#251

cross311:
Looks like @didi0613 is moving ahead with this!  @DavidOrchard

didi0613:
@cross311  Yes i am currently working on updating archetype component.
eslint part has been merged & currently working on upgrading webpack v1 to v2

animesh10:
closing since https://github.com/electrode-io/electrode/pull/251 is merged.

PR #237, opened 2017-03-14
Using library versions from component

cross311:
Given the current order, if a project using the component archetype then the versions from the archetype win.  This is an issue, especially when using latest eslint configurations.  in the archetype eslint 1.*  wins, which is super old and does not support string severity.

### DIFF
--- a/packages/electrode-archetype-react-component/archetype-gulpfile.js
+++ b/packages/electrode-archetype-react-component/archetype-gulpfile.js
@@ -8,9 +8,9 @@ const exec = gulpHelper.exec;
 
 function setupPath() {
   const nmBin = "node_modules/.bin";
-  gulpHelper.envPath.addToFront(Path.resolve(nmBin));
-  gulpHelper.envPath.addToFront(Path.join(archetype.devPath, nmBin));
   gulpHelper.envPath.addToFront(Path.join(__dirname, nmBin));
+  gulpHelper.envPath.addToFront(Path.join(archetype.devPath, nmBin));
+  gulpHelper.envPath.addToFront(Path.resolve(nmBin));
 }
 
 function setProductionEnv() {


cross311:
This person also ran into it #227

jchip:
but then this flips the situation.  You may be invoking a command that's not compatible with what the archetype needs.

cross311:
@jchip So are you concerned that newer versions are going to break the archetypes?  This is a blocker for running the react standard eslints from airbnb.  This order of loading is very confusing, it took me 2-3 hours just to figure out why eslint v1 was being used.

I understand what you are saying but electrode is not going to be able to keep up with package updates, so locking the versions to first use what is in the archetypes, is continue going to come up as blockers.

cross311:
The dev path should be a fallback.  The dev npm package currently is used for an easy way to make sure you dont have to include all these packages into the main repository so the team can update them.  How do you recommend another company that wants to use newer version of things gets stuff in?  Do we need to generate our own dev package?

cross311:
creating our own dev package does not look like an option since the name of the dev package is hardcoded.

jchip:
The issue came up for us at some point due to a 3rd party package pulling in a diff version of a command the archetype uses and wins during npm install.  The idea of the archetype is we want it to be the primarily setups for the component so it's reasonable to make sure everything comes from it first.

For situation where user wants to deviate from certain setup the archetype needs to allow overrides.  In this instance seems we need to allow commands to be override.

cross311:
looking at some other examples, people are just not using the archtypes gulp check, they are not using it and just doing npm scripts:

`"check": "npm run test && npm run lint && npm run flow",`

So looks like others using the OSS are having this issue as well.  @ananavati 's example is where i am pulling this from

jchip:
Yeah, the component archetypes have not been kept up to date for a while now.  We have a backlog item to bring it up to date.

cross311:
I unblocked my team by using the example, I posted.  How can I help with what you suggested above so that we can switch back eventually to the gulp check.  I maybe pass the dev package name via archetype gulp function params?

jchip:
OK, we've not had much resource dedicated to the component archetype for a while now.  We are hoping to do some update on it soon.  After that, then add some of the customization support we added to the app archetype.   As for changing the path orders, I think it's best to avoid that for now since that could alter some other behaviors.

cross311:
I am going to close this because the implications could be big.

Issue #236, opened 2017-03-14
Yarn Incompatible Engines


cross311:
`error electrode-archetype-react-component@1.1.12: The engine "node" is incompatible with this module. Expected version "^4.2.6".`

Is there a reason why the npm modules specify an engine?  Can that be removed so that using yarn does not default throw and error?

cross311:
@jchip I can remove all the engine definitions, unless there is a real reason for them?

cross311:
cool!

PR #241, opened 2017-03-14
Removed engines from package.json of archetype-react-component

jbarabander:
As mentioned in #236.  Attempting to add electrode-archetype-react-component as a dependency in another project causes yarn to throw errors by default unless the version of node being used is v4.x.  Removing the engines portion of the package.json fixes the issue.

This error can be circumvented of course by running 
```yarn install --ignore-engines``` 
but when working with a library like lerna this is not currently possible.

### CODE REVIEW

**Comment by jbarabander** [comment]
@cross311

**Comment by jchip** [comment]
how about add the other versions to the engine section?

**Comment by jbarabander** [comment]
@jchip: Sure thing!

**Comment by jchip** [comment]
btw, the engines section was added because the archetypes require npm >= 3.

**Comment by jbarabander** [comment]
Gotcha.  so would it be ok to change it to this instead then?:
```
    "engines": {
         "node": ">=4",
         "npm": ">=3"
    }
```

jchip:
yup, that should work.  Obviously some future versions of node/npm could be breaking though, but I don't think that's a problem.

jbarabander:
@jchip: Good point.  In light of that I opted for doing what was in the electrode-archetype-react-app and just added the additional version numbers to the engines piece.

Issue #190, opened 2017-02-17
CSS Module Hook: Flash of Unstyled Content


seansean11:
We're currently writing CSS for an Electrode app that takes advantage of the included postcss webpack plugins.  For example, we use quite a few [postcss-import](https://github.com/postcss/postcss-import) `@import` statements  and a lot of simple nesting, provided by [postcss-cssnext](https://github.com/MoOx/postcss-cssnext)
```css
.Page {
  &_content {
    ...
  }
}
```
I recently noticed that there is a flash of unstyled content at the beginning of every page load.  This seems to be due to the fact that css-modules-require-hook on the server doesn't know how to handle the postcss syntax with its default setup in `server->index.js`.  
```js
supports.cssModuleHook({
  generateScopedName: "[name]__[local]___[hash:base64:5]"
});
```
The HTML delivered from the server does not have any classnames for nested classes, so they are generated on the client.  For the split second before they are generated, you see a flash of unstyled content.

I tried supplying the postcss plugins directly to css-modules-require-hook, as I saw done in their documentation, and also found out it only accepts an [older version](https://github.com/rxlabs/tasty-pokedex/issues/11) of postcss-import. 
```js
supports.cssModuleHook({
  use: [
    require("postcss-import"),
    require("postcss-cssnext")
  ]
});
```
When running the app with this setup, I get no errors, but the class attributes are still not getting generated on the server-side, rather appearing after the html loads on the client.

Has anyone else dealt with this and figured out a solution?  For the time being I'm wrapping the problematic unstyled content with the [above-the-fold-only-server-render](https://github.com/electrode-io/above-the-fold-only-server-render) and not rendering it on the server altogether.

citelao:
Is this specific to `@import`? I am having the same problem.

seansean11:
@citelao we noticed the issue first with nested selectors not getting class names generated on the server.
```css
.Page {
  &_content {
    ...
  }
}
```
I then looked into the css-modules-require-hook and it seems as though we should be registering our postcss-cssnext plugin here, or else it won't know how to interpret the nested syntax for SSR.  After registering the postcss-cssnext plugin like in my above comment, the server actually displayed errors saying that it didn't recognize variable names that we were using (they were being imported).  Since the CSS webpack is using postcss-import I added that to my CSS module hook as well.  When I did this, the errors disappeared (and I thought everything might work), but the class names still did not generate on the server, leaving us with the same flash of unstyled content issue.

[...]

PR #253, opened 2017-03-22
Fix CSS modules class name

karlhorky:
Closes electrode-io/electrode#190


karlhorky:
I deleted the original repo and reforked, so opening a new PR for this: #258

PR #268, opened 2017-04-10
Accept Parameters to check if the Generator has been extended

animesh10:
Accepts and checks for parameters for a `gitHubUrl` and a flag `isExtended` and `serverType`.
The generator that is using the electrode-OSS generator can now pass these values to the OSS generator.
The `githubUrl` will override the repo url for `package.json` and the .`git`.
`serverType` with set the props and will hide the prompt asking for the server type. This helps control the kind of server you want to use in your project.
`isExtended` is currently only used to delay the dependency installation and we transfer control back to the calling generator to complete the installation.

### DIFF
--- a/packages/generator-electrode/generators/app/index.js
+++ b/packages/generator-electrode/generators/app/index.js
@@ -57,15 +57,9 @@ module.exports = generators.Base.extend({
       desc: 'Content to insert in the README.md file'
     });
     //Flag to check if the OSS generator is being called as a subgenerator
-    if (this.options && this.options.isExtended) {
-      this.isExtended = true;
-    }
-    if (this.options && this.options.serverType) {
-      this.serverType = this.options.serverType;
-    }
-    if (this.options && this.options.githubUrl) {
-      this.githubUrl = this.options.githubUrl;
-    }
+    this.isExtended = this.options.isExtended || false;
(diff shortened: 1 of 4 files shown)


packages/generator-electrode/generators/app/index.js
```diff
@@ -90,6 +99,9 @@ module.exports = generators.Base.extend({
       this.props.authorEmail = info.email;
       this.props.authorUrl = info.url;
     }
+    if (this.serverType) {
+      this.props.serverType = this.serverType || HapiJS;
```
jchip: redundant checking `this.serverType`

packages/generator-electrode/generators/app/index.js
```diff
@@ -57,6 +56,16 @@ module.exports = generators.Base.extend({
...
     });
+    //Flag to check if the OSS generator is being called as a subgenerator
+    if (this.options && this.options.isExtended) {
+      this.isExtended = true;
+    }
+    if (this.options && this.options.serverType) {
+      this.serverType = this.options.serverType;
+    }
+    if (this.options && this.options.githubUrl) {
+      this.githubUrl = this.options.githubUrl;
```
jchip: why do you have to check each one?  Why not just directly assign each?  You are checking them every where anyways.

PR #269, opened 2017-04-11
Add fonts to head

kaytcat:
We need a way to add our font to the head bundle so I've added a config key called fontPath where you can specify your font url and if it is defined it will add it to the head.


kaytcat:
Sorry.. I meant to open this with our own fork

PR #274, opened 2017-04-12
Add react test dependencies to project to meet enzyme requirements

seansean11:
**Context**
As soon as Enzyme released [V2.8.1](https://github.com/airbnb/enzyme/releases/tag/v2.8.1) earlier today I noticed that the gulp-check command failed with an error message that stated "react-test-renderer is an implicit dependency in order to support react@15.5+.  Please add the appropriate version to your devDependencies.  See https://github.com/airbnb/enzyme#installation"

**Solution**
I can't tell you why this started to throw errors when Enzyme moved from V2.8.0 to V2.8.1, but in the documentation it clearly states that `react-test-renderer` and `react-dom` should also be installed with it for use with React versions over 5.5, and installing them seems to remedy the error.

package_json_-_discover_medical_dynamichealthui


jchip:
thanks for the PR.  I already added the dep.

jchip:
closing.  already done in https://github.com/electrode-io/electrode/pull/273

seansean11eis:
@jchip , thanks

PR #280, opened 2017-04-14
universal-material-ui: [Chore][Major] Update Material-UI sample to src/lib mode

animesh10:
material-ui sample app migration to `src-lib` mode.
- move `client` and `server` to `src` folder.
- add `babelrc` for client and server
- update `config` and `package.json` to point to correct `main`.
- update `src/server/index.js` to **not** use `babel-register`


didi0613:
lgtm 👍

jchip:
you would have to update the sample to src/lib mode also.

Issue #97, opened 2016-12-28
Webpack config customization


sbuys:
Are there any pathways to customizing the webpack config? e.g. adding loaders

vimcat:
Now that #109 has been merged, how do we use it? Do we just have to add a webpack object into config/default.js?

nevos:
Still trying to figure out how to customize my host. I've added 
    "publicPath": "http://MY_IP:2992/js/" 
to my entry.config.js
but my .isomorphic-loader-config.json is overriding that.

ananavati:
@vimcat 

pr #109 has been merged, pending:
1. writing up `Usage` docs in README.
2. publishing a new version of the `electrode-archetype-react-app` with this change (currently targeted for 01/31).

tl;dr on how to extend/override webpack config:

Everything in [config/webpack](https://github.com/electrode-io/electrode/tree/master/packages/electrode-archetype-react-app/config/webpack) that starts with `webpack` is extendable by putting a file of the same name in your root. So `webpack.config.js, webpack.config.test.js`, etc..

for example: @jshado1 brought up a use-case here https://gitter.im/electrode-io/electrode

Use-case: `how to register with Electrode a postcss plugin, like postcss-strip-inline-comments`

Solve: 
define a file : `/<app>/webpack.config.js` with this as the code inside of it:
```
const baseConfig = require("electrode-archetype-react-app/config/webpack/webpack.config.js");

module.exports = {
    postcss: function() {
        const plugins = baseConfig.postcss();

        return [...plugins, 
            // new plugins go here
        ]
    }
};
```

citelao:
@ananavati 's method is working; is this issue waiting on documentation?

karlhorky:
> writing up Usage docs in README

Does this still need to be done?

cc @jmcriffey @ananavati @jchip

karlhorky:
Actually, I've just opened #262 for this.

karlhorky:
@ananavati you could link directly to the point in history in gitter about the original use case:

https://gitter.im/electrode-io/electrode?at=588156d56c1635643c5058bd

PR #284, opened 2017-04-17
Add readme for extending webpack config

animesh10:
Added how the folder should look like and a sample webpack config.


karlhorky:
It looks like this `{% raw %}` directive doesn't seem to work in GitHub Flavored Markdown:

screen shot 2017-04-18 at 11 10 12

https://github.com/animesh10/electrode/blob/9ff25e26741ac972cc855da39abb5730204107ee/packages/electrode-archetype-react-app/README.md

Maybe a code block instead?

packages/electrode-archetype-react-app/README.md
```diff
@@ -236,6 +236,35 @@ var supports = require("electrode-archetype-react-app/supports");
...
+var BellOnBundlerErrorPlugin = require('bell-on-bundler-error-plugin')
+
+const config = {
+  plugins: [
+    new BellOnBundlerErrorPlugin()
+  ]
+};
+
+module.exports = config;
+};
```
karlhorky: I guess we don't need this curly brace and semicolon?

packages/electrode-archetype-react-app/README.md
```diff
@@ -236,6 +236,35 @@ var supports = require("electrode-archetype-react-app/supports");
...
+└── config
+    └── webpack
+        └── webpack.config.js
+
+{% endraw %}
+
+Where `webpack.config.js` is your extended or overriding webpack config, for example:
+
+```js
+var BellOnBundlerErrorPlugin = require('bell-on-bundler-error-plugin')
```
karlhorky: Do we want semicolons everywhere?

karlhorky: Should this be a `const` to be consistent with the code style below?

packages/electrode-archetype-react-app/README.md
```diff
@@ -236,6 +236,35 @@ var supports = require("electrode-archetype-react-app/supports");
 
 supports.cssModuleHook();
 ```
+## Extending Webpack Configuration
+
+The webpack config for your app is being generated by the archetype. While composing the webpack config, the archetype looks for the `archetype/config` folder at the project root of your app.
+In order to override or extend the webpack configuration you should create a file under `archetype/config/webpack`.
+Your directory structure should look like this:
```
karlhorky: Extra empty lines between these three lines would make them into separate paragraphs, which would help with readability:

```diff
The webpack config for your app is being generated by the archetype. While composing the webpack config, the archetype looks for the `archetype/config` folder at the project root of your app.
+                                                                                                    
In order to override or extend the webpack configuration you should create a file under `archetype/config/webpack`.
+                                                                                                    
Your directory structure should look like this: 
```

packages/electrode-archetype-react-app/README.md
```diff
@@ -236,6 +236,35 @@ var supports = require("electrode-archetype-react-app/supports");
 
 supports.cssModuleHook();
 ```
+## Extending Webpack Configuration
+
+The webpack config for your app is being generated by the archetype. While composing the webpack config, the archetype looks for the `archetype/config` folder at the project root of your app.
+In order to override or extend the webpack configuration you should create a file under `archetype/config/webpack`.
```
karlhorky: Is this path `archetype/config/webpack` actually correct?

In https://github.com/electrode-io/electrode/issues/97#issuecomment-274150029 @ananavati mentioned that webpack configuration files in the root of the app will be used to extend the configuration. I've tested this to be working 2-3 weeks ago.

Has this been changed?

animesh10: I think this change might have occurred around the same time, the way mentioned in the comment should still work, but this is the way for extending your webpack config, https://github.com/electrode-io/electrode/blob/master/packages/electrode-archetype-react-app-dev/config/webpack/util/generate-config.js#L15

karlhorky: Oh nice, they both work! Do we want to mention both in the readme?

seansean11eis:
@karlhorky thanks for adding this to the documentation.  Reading the docs, I'm still left wondering how one might override current webpack configuration.  Electrode is fairly opinionated and I think it is very likely that users will want to override settings in webpack (i.e. currently trying to turn off css modules).  I think that this section could be augmented.  Is there anywhere that you can find config override info?  Or should I be looking at the webpack-config-composer project.

karlhorky:
@seansean11 Sorry that this is still unclear. In your project you would:

1. Add a folders structure like: `<PROJECT ROOT>/archetype/config/webpack`
2. Create a webpack configuration file matching the name of a file [here](https://github.com/electrode-io/electrode/tree/65884aa271ed76398c7bc062675d99627da94b8e/packages/electrode-archetype-react-app-dev/config/webpack) that you want to override (for instance, `webpack.config.dev.js` or `webpack.config.js`)
3. In this file, include the minimal configuration to override the key in the configuration that you want to affect. @ananavati mentions an example in https://github.com/electrode-io/electrode/issues/97#issuecomment-274150029. This will be merged by `webpack-config-composer`.

Does this explanation help?

seansean11eis:
Hey @karlhorky thanks so much for the response.  I actually tried following the steps in @ananavati's comments before posting.  The electrode archetype has changed a bit since that comment.  I import the default settings like so:
``` js
const baseConfig = require("electrode-archetype-react-app-dev/config/webpack/webpack.config.js");
```
This require returns an empty object, so I'm a little confused on how to obtain default settings to override.

jchip:
https://docs.electrode.io/chapter1/intermediate/app-archetype/webpack-config.html#overriding

Issue #92, opened 2016-12-21
[Windows] Code Coverage not being generated


alecwhittington:
When running `gulp check` on Windows 10, code coverage is not being generated properly. Linting is working correctly as are the tests running. It appears the code coverage files are being generated (based on the file stamps as well as the message in the index.html file that is created).

On my Mint Linux machine, the code coverage is working as expected, so I am pretty sure this is just a Windows only issue.

No errors are thrown, the code coverage table in the command line shows 100% coverage, but no files. This is also what is shown in the Index.html. When run on Mint, all files are listed in both locations and code coverage shows that all are 100% except Branches which shows approximately 96%.

**Edited to add more information

alecwhittington:
Been able to narrow it down to Karma not writing anything to the coverage.json file on windows. Have not been able to narrow down as to why yet.

Issue #80, opened 2016-12-17
Windows 10 - Cannot run Gulp Check due to Sinon errors


alecwhittington:
I've been trying to get our tests running from a Generated Component project, but have not been able to get any further than linting as Sinon is causing problems.

Here is what I am seeing from my console (truncated to show only the important pieces):

17 12 2016 12:32:23.618:INFO [karma]: Karma v0.13.22 server started at http://localhost:8080/
17 12 2016 12:32:23.634:INFO [launcher]: Starting browser PhantomJS
17 12 2016 12:32:24.044:INFO [PhantomJS 1.9.8 (Windows 8 0.0.0)]: Connected on socket akvOIblAU_ceFTgQAAAA with id 21202378
PhantomJS 1.9.8 (Windows 8 0.0.0) ERROR
  Error: Cannot find module './sinon/util/core'.
  at F:/@Salucro/salucro-reactor-common/node_modules/electrode-archetype-react-component/config/karma/entry.js:49021

PhantomJS 1.9.8 (Windows 8 0.0.0): Executed 0 of 0 ERROR (0.094 secs / 0 secs)

Unfortunately this is a blocking issue for us as it makes zero sense to continue writing components when we cannot test them through the component archetypes. 

I've spent more than an hour researching the issue, but cannot get past the error. I've verified that the node_modules/sinon/lib/sinon/util folder exists and there is a core module there. Seems like some sort of pathing issue based on the cannot find module error.

Also tried generating another component project and just running gulp check as soon as it is generated and NPM modules are installed. Same result

alecwhittington:
More information:

Found the cause of the error. In the node_modules\electrode-archetype-react-component\config\webpack\webpack.config.test.js file, line 24. The pathing will not work for Windows.

Had to change from:
```javascript
/node_modules\/sinon\//
```
to
```javascript
/node_modules\\sinon\\/
```

for it to work on windows. This is a pretty big problem as we have developers that use Windows, Mac, and Linux.

alecwhittington:
Fixed issue for Windows as well as Linux bases systems. Pull Request issued

jchip:
thanks for the fix. published https://github.com/electrode-io/electrode/commit/1caca411b6de686b79ec747c3c89d4af0b41b4a3

PR #292, opened 2017-04-19
Fix Windows coverage errors and Windows Sinon error

seansean11:
**Tested on Windows 10 and Mac OS10.12.4**

**Context**
Cannot run `npm test` on Electrode apps on Windows devices without these to errors:
1. #92 Istanbul coverage failing on windows
2. #80 Was applied to the component archetype but not the app archetype

**Fixes**
1. Fix windows coverage errors #92 with solution https://github.com/gotwarlost/istanbul/issues/90
2. Determine regex value based on whether the device platform is Windows. (#82)

**I would really like to apply this fix to the newest 1.x version because we are currently using 1.10.0 in production and can't run our tests on any Windows devices.  We can't update to 2.x yet because there are quite a few known bugs.  Should 1.* updates be made to the webpack-1.0 branch?*

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/test-base.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/test-base.js
@@ -1,5 +1,9 @@
 "use strict";
 
+const sinonRegex = process.platform === "win32" ?
+  /node_modules\\sinon\\/ :
+  /node_modules\/sinon\//;
+
 module.exports = {
   module: {
     /*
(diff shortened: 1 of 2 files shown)


packages/electrode-archetype-react-app-dev/config/webpack/partial/test-base.js
```diff
@@ -10,7 +14,7 @@ module.exports = {
      * https://github.com/sinonjs/sinon/pull/600#issuecomment-162529457
      */
     noParse: [
-      /node_modules\/sinon\//
```
jchip:
```js
const Path = require("path");

new RegExp(Path.normalize("node_modules/sinon/").replace(/\\/g, "\\\\"))
```

seansean11: thanks for the tip 👍

PR #296, opened 2017-04-20
Fix CSS Modules FOUC with production build

kaytcat:
The production build uses the `lib` folder for built assets. Because CSS
Modules uses the file path to generate the unique hash for the class names,
this causes an inconsistency between the client and server class names, which
causes a Flash Of Unstyled Content (FOUC) in production mode.

cc @karlhorky


karlhorky:
I guess we'll want to detect for something that the npm `prod` script sets (or even something the server sets) - because currently it doesn't set the `NODE_ENV` to `'production'`.

PR #304, opened 2017-04-24
Fix deps bug

kaytcat:
This was causing my previous fouc fix to break since it was installing the incorrect version


kaytcat:
Sorry meant to open with fork

PR #305, opened 2017-04-26
[WIP] invalid host header fixes for oss archetype

didi0613:
Same fixes for invalid host header for oss app archetype


packages/electrode-archetype-react-app/arch-gulpfile.js
```diff
@@ -42,6 +42,8 @@ function setStaticFilesEnv() {
 
 function setWebpackDev() {
   process.env.WEBPACK_DEV = "true";
+  process.env.WEBPACK_DEV_PORT = 3000;
+  process.env.WEBPACK_DEV_HOST = "dev.walmart.com";
```
jmcriffey: I think this is going to need to be configured by the user.

PR #313, opened 2017-05-02
archetype-react-component: [Bug] [Patch] Clean remaining dev and hot tasks

animesh10:
Archetype component still referred to tasks that were previously deleted, i.e. `dev` and `hot`.


didi0613:
LGTM 👍

PR #315, opened 2017-05-03
archetype-react-app: [bug] [Patch] Update initial entry path for PWA

animesh10:
The initial value of entry can be empty which was causing webpack validation errors, causing webpack compilation to fail.


didi0613:
LGTM 👍

PR #318, opened 2017-05-04
generator-electrode: [bug][patch] Accept props as options

animesh10:
The `component-add` generator can accept and initialize props if they are passed to it.


didi0613:
lgtm 🎉

PR #320, opened 2017-05-05
Fix help message

dutradda:
Added the serverType option on help


animesh10:
The prompt message is there already: https://github.com/electrode-io/electrode/blob/master/packages/generator-electrode/generators/app/index.js#L135-L140

dutradda:
@animesh10 how to show the prompt? I ran the generator many times and this option just showed on versions prior 1.9..

animesh10:
Yes, that was a bug that got introduced in that release. Thanks for noticing and raising the PR.
It has been fixed and will be available in the next release.

animesh10:
Closing, since fix has been published.

PR #323, opened 2017-05-06
Universal-react-node: [bug][path] Update Path for Dockerfile

animesh10:
The dockerfile for the sample app still points to the old server location.


didi0613:
lgtm 👍

PR #353, opened 2017-06-01
[patch][bug-fix][app-archetype] FIX gulp lint as it skips src/server files

didi0613:
FIX gulp lint as it skips src/server files (works fine on master).
Home aoo unifying oss archetype required.


ajoshi0:
LGTM @didi0613

Issue #359, opened 2017-06-06
Allow production .js and .css bundles to be located on a CDN


bodawei:
We are using electrode-react-webapp (along with the rest of electrode) but need to have the main .js file (and .css file) be located on a CDN when in production.

My experience, and reading of the code in react-webapp.js suggests that this is currently not possible with electrode-web-app.

### The Situation
In: https://github.com/electrode-io/electrode/blob/master/packages/electrode-react-webapp/lib/react-webapp.js ,  `BODY_BUNDLE_MARKER` is filed in with `<script src="${js}"></script>` where js comes from 
```
    const bundleJs = () => {
      if (!renderJs) {
        return "";
      }
      return WEBPACK_DEV ? devJSBundle : jsChunk && `/js/${jsChunk.name}` || "";
    };
```

So, unless I'm overlooking something...  in production mode, the url for the base javascript bundle is always going to be `<script src="/js/(name)"></script>`, but we need `<script src="http://awesome-cdn.com/js/(name)"></script>`

I have seem the references to this page  https://github.com/electrode-io/isomorphic-loader#usage-with-cdn-server and what it says about changing the publicPath in the webpack config.  Yet, this only seems applicable to running in dev mode with webpack-dev-server.

### Proposed solution
What I'm thinking here is to provide a pull request which would define an additional plugin option (configurable from confippet, and defaulting to '' in `react-webapp.js`) which could be used to provide a base URL to be prepended to the bundle url suffix in `makeBodyBundles()`.

If this sounds good, I can put together a change and make a pull request.  Thoughts?

PR #360, opened 2017-06-06
react-web-app: [feat] [minor] Allow an external base URL for resources on production runs (#359)

bodawei:
This is a proposed fix for https://github.com/electrode-io/electrode/issues/359

Note: I am unsure whether to categorize this as "feat" or "bug".  It seems like a feature to me, but it is so tiny that it hardly seems to justify that designation!

Details below.

### Problem
In some cases, it is desirable to have the main JavaScript, CSS and pwa manifests reside on a CDN server while running in production mode.  This is not possible given the current implementation of `electrode-react-webapp`.

### Solution
This adds a new configuration option, `prodBundleBase`, which defaults to "", but can be configured in the confippet configuration system for plugins.

When specified, this will be prefixed to the url fragment for the main JavaScript bundle(s), css files and pwa manifest files.

### Implementation notes
This adds several tests to validate that all the URL's are properly updated.

Note that there are two tests, not in react-web-app, that are failing.  These fail in a "vanilla" checkout.

### DIFF
--- a/packages/electrode-react-webapp/lib/react-webapp.js
+++ b/packages/electrode-react-webapp/lib/react-webapp.js
@@ -152,14 +152,14 @@ function makeRouteHandler(routeOptions, userContent) {
     );
 
     const bundleCss = () => {
-      return WEBPACK_DEV ? devCSSBundle : cssChunk && `${prodBundleBase}/js/${cssChunk.name}` || "";
+      return WEBPACK_DEV ? devCSSBundle : cssChunk && `${prodBundleBase}${cssChunk.name}` || "";
     };
 
     const bundleJs = () => {
       if (!renderJs) {
         return "";
       }
-      return WEBPACK_DEV ? devJSBundle : jsChunk && `${prodBundleBase}/js/${jsChunk.name}` || "";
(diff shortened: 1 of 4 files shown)


packages/electrode-react-webapp/lib/react-webapp.js
```diff
@@ -151,22 +152,23 @@ function makeRouteHandler(routeOptions, userContent) {
     );
 
     const bundleCss = () => {
-      return WEBPACK_DEV ? devCSSBundle : cssChunk && `/js/${cssChunk.name}` || "";
+      return WEBPACK_DEV ? devCSSBundle : cssChunk && `${prodBundleBase}/js/${cssChunk.name}` || "";
```
jchip: the `/js` part should come from the bundle config as well, but default it to `/js`.

jchip:
thanks for the PR.  could you also update the README here https://github.com/electrode-io/electrode/tree/master/packages/electrode-react-webapp#options please?

packages/electrode-react-webapp/lib/react-webapp.js
bodawei: Good call!  Definitely will do this first thing on Monday.

(and good call on the README.  Ironically enough, I'd gone looking for something like that, but somehow didn't think of the README itself. Very grateful to you for pointing it out).

PR #368, opened 2017-06-13
archetype-react-component: [chore] [patch] update travis command for new component generator

animesh10:


jchip:
individual package's CI file no longer has any effect.  should just remove them.

PR #374, opened 2017-06-14
generator-electrode: [patch] [chore] Set ES6 flag in gulpfile.

animesh10:
- set ES6 test flag that is used by the `arch-gulpfile`.
- `test/server` eslint file extends from the new es6 server test file.

### DIFF
--- a/packages/generator-electrode/generators/app/templates/gulpfile.js
+++ b/packages/generator-electrode/generators/app/templates/gulpfile.js
@@ -1,2 +1,2 @@
-process.env.ES6_LINT = true;
+process.env.SERVER_ES6 = true;
 require("electrode-archetype-react-app")();
--- a/packages/generator-electrode/generators/app/templates/gulpfile.js
+++ b/packages/generator-electrode/generators/app/templates/gulpfile.js
@@ -1,1 +1,2 @@
+process.env.ES6_LINT = true;
 require("electrode-archetype-react-app")();
--- a/packages/generator-electrode/generators/app/templates/test/server/eslintrc
+++ b/packages/generator-electrode/generators/app/templates/test/server/eslintrc
@@ -1,6 +1,6 @@
 ---
 extends:
-  - "../../node_modules/electrode-archetype-react-app-dev/config/eslint/.eslintrc-mocha-test"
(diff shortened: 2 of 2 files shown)


packages/generator-electrode/generators/app/templates/test/server/eslintrc
```diff
@@ -1,6 +1,6 @@
 ---
```
jchip: where is this file being copy in the code?

packages/generator-electrode/generators/app/templates/gulpfile.js
```diff
@@ -1 +1,2 @@
+process.env.ES6_LINT = true;
```
jchip: `process.env.SERVER_ES6`

Doesn't the archetype need to look at this flag and decide to use ES6 eslint to lint the code in `src/server`?

packages/generator-electrode/generators/app/templates/test/server/eslintrc
animesh10: the entire test folder is being copied [here](https://github.com/electrode-io/electrode/blob/master/packages/generator-electrode/generators/app/index.js#L329). Then eslint files moved [here](https://github.com/electrode-io/electrode/blob/master/packages/generator-electrode/generators/app/index.js#L378)

packages/generator-electrode/generators/app/templates/gulpfile.js
animesh10: The `lint-server` task is using  [.eslintrc-node](https://github.com/electrode-io/electrode/blob/master/packages/electrode-archetype-react-app-dev/config/eslint/.eslintrc-node) which in turn is extending [es6-node](https://github.com/electrode-io/electrode/blob/master/packages/electrode-archetype-react-app-dev/config/eslint/.eslintrc-node#L3)

jchip: then what's the point of setting this?

animesh10: The task `lint-server-test` is using `.eslintrc-mocha-test` which is extending `es5-node` configurations. 
This flag will switch the new generated apps to use the `eslintrc-mocha-test-es6` config which extends `es6-node` config for the `lint-server-test` task.

PR #375, opened 2017-06-14
electrode-archetype-react-app-dev: [patch][chore] Allow webpack config to override test entry point

Nickersona:
My goal here is to co-locate my test files with my components. Something I thought was possible with electrode but it must have changed at some point.  My solution is to make the test entry file overridable via the Webpack config. It's worked for my case, and gives me a ton of control over my testing environment which is very valuable to my project.  

Based on the naming of the webpack parital `test-entry.js` my assumption is that this was the intention of the webpack config file, but it doesn't look like it's being applied properly in the `karma.conf.js`.

Is there a reason that the test entry file is hard coded? This is quite limiting, but if there's an alternative way you'd suggest overriding the entry file I'd be interested to hear the approach.

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/karma/karma.conf.js
+++ b/packages/electrode-archetype-react-app-dev/config/karma/karma.conf.js
@@ -4,7 +4,7 @@ const Path = require("path");
 
 const webpackCfg = require("../webpack/webpack.config.test");
 
-const MAIN_PATH = require.resolve(webpackCfg.entry);
+const MAIN_PATH = webpackCfg.entry;
 
 const PREPROCESSORS = {};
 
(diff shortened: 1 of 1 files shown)


jchip:
Thanks for the PR.  Yes, the change makes sense.  Could you rebase to the latest please?

Nickersona:
PR updated with the latest from master

Nickersona:
@jchip Do you know why the tests would fail for this. I'm struggling to get them running locally on master. I'm not sure why they're failing

jchip:
sorry, I will have to look into this later.

jchip:
Hi, just looking this a bit.  Webpack config's entry points to the app's main JS file, usually `src/client/app.jsx`.  Also, it's not necessary a plain string.  It could be an object with multiple entry for webpack.  Karma's entry has to point to its own `entry.js`, which does bootstrapping to get the React app loaded into karma.  

I am not sure how this PR is supposed to work actually.

Nickersona:
Thanks for looking into this. My approach was wrong here. I was able to accomplish what I wanted through a custom test Entry file. Closing the PR out.

PR #423, opened 2017-06-22
Move react-helmet scripts to head to avoid double load

sergesemashko:
This PR moves all react-helmet scripts to head because they are being loaded/rendered twice if they are [not placed in head tag](https://github.com/nfl/react-helmet/issues/149#issuecomment-292467049)


sergesemashko:
thanks, I'll take a look today.

sergesemashko:
sorry for the delay, @jchip, eslint warning is fixed now

jchip:
thank you.

sergesemashko:
@jchip, no problem. 
I'm looking [intex.html](https://github.com/electrode-io/electrode/blob/master/packages/electrode-react-webapp/lib/index.html#L9) and can get why `{{PREFETCH_BUNDLES}}` is placed in the `<head>`? It seems like serialized redux state won't be needed until `<script>if (window.webappStart) webappStart();</script>` [on line 15 of index.html](https://github.com/electrode-io/electrode/blob/master/packages/electrode-react-webapp/lib/index.html#L15) and moving it later would speed up the loading and parsing of critical pieces on the page.

I'm thinking to submit another PR to fix that, but maybe there is a reason why it is in `<head>`?

PR #445, opened 2017-06-29
[patch] [bug] No git init for demo-app, correct default componentName.

animesh10:
- app generator: do not call git generator for the demo-app.
- add component generator: correct default for component name.

### DIFF
--- a/packages/generator-electrode/generators/app/index.js
+++ b/packages/generator-electrode/generators/app/index.js
@@ -427,20 +427,22 @@ module.exports = generators.Base.extend({
         local: require.resolve("../editorconfig")
       }
     );
-
-    this.composeWith(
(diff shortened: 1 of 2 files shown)


packages/generator-electrode/generators/app/index.js
```diff
@@ -427,20 +427,22 @@ module.exports = generators.Base.extend({
...
     );
-
-    this.composeWith(
-      "electrode:git",
-      {
-        options: {
-          name: this.props.name,
-          githubAccount: this.props.githubAccount,
-          githubUrl: this.props.githubUrl
+    if (!this.isDemoApp) {
```
jchip: make the name more generic such as `initGit`

PR #452, opened 2017-07-02
[patch][feat] Component archetype chrome headless

didi0613:


packages/electrode-archetype-react-component-dev/config/karma/karma.conf.dev.js
```diff
@@ -26,5 +24,16 @@ module.exports = function (config) {
         ui: "bdd"
       }
     }
-  });
+  };
+
+  if (browser === "chrome") {
```
jchip: please put this code in a common file

didi0613: @jchip will do

PR #476, opened 2017-07-13
[patch] [chore] removed vscode plugin settings

donsalari:
Also, rearranged package.json from most frequently needed information to least.


animesh10:
👍

PR #479, opened 2017-07-16
Custom html markers proposed solution

sergesemashko:
This PR just shows the prove of concept for having custom HTML markers. 

**Motivation**

There are some cases (mostly side effects) requiring custom placeholder to render dynamic content. It's very tricky to add something on the page after `React.renderToString()`.

Examples, where custom markers would be very handy: 
- custom [react-side-effect](https://github.com/gaearon/react-side-effect) modules
- store dependent scripts
- any sort of dynamic code that doesn't belong to React

**Solution**

It turned out that it's relatively easy to add custom markers functionality to the electrode rendering workflow.

It makes sense to hear any concerns or opinions how to do it better.

### DIFF
--- a/packages/electrode-react-webapp/lib/react-webapp.js
+++ b/packages/electrode-react-webapp/lib/react-webapp.js
@@ -161,6 +161,9 @@ function makeRouteHandler(routeOptions, userContent) {
           case CRITICAL_CSS_MARKER:
             return criticalCSS;
           default:
+            if (content.markers && content.markers[m] !== undefined) {
+              return content;
+            }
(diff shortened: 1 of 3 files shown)


jchip:
Thanks for the proposal and PR. 

We've had a solution for a while to support arbitrary custom markers.

David is working on implementing the solution here https://github.com/dvonlehman/electrode/commit/21381cf0cc1d8ec716204a20d3214c9edcf97a2b and I believe he's almost done.

sergesemashko:
@jchip, awesome, looking forward to get it in. Closing this PR

PR #480, opened 2017-07-16
Move {{PREFETCH_BUNDLES}} marker after {{SSR_CONTENT}}

sergesemashko:
**Motivation**

It's slightly unclear why serialized state is placed into `<head>`. The store state might be pretty big sometimes and it looks like it's more critical to parse CSS and page HTML first. 

It would be great to know if there is a strong reason, and possibly make it more obvious why {{PREFETCH_BUNDLES}} is placed before {{SSR_CONTENT}}.

### DIFF
--- a/packages/electrode-react-webapp/lib/index.html
+++ b/packages/electrode-react-webapp/lib/index.html
@@ -6,11 +6,11 @@
     {{META_TAGS}}
     {{PAGE_TITLE}}
     {{CRITICAL_CSS}}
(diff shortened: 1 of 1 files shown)


jchip:
there's no particular reason technically.  Internally at WML we have a edge side caching layer that does sophisticated parsing of the HTML passing thru it and there was some requirement to have the hydrated store in HEAD.

jchip:
not merging yet.  I will try it out with the samples first.

PR #481, opened 2017-07-16
Redux state should be preloaded after renderToString.

khayong:
If a component have a dispatch call to update state, the `window.__PRELOADED_STATE__` does not prefetch with the updated state.

Component could call the store dispatch at
- Constructor
```
class Test extends React.Component {
  constructor(props) {
    super(props);
    store.dispatch(); // Call to update state.
  }
}
```
- Component will mount
```
class Test extends React.Component {
  componentWillMount() {
    store.dispatch(); // Call to update state.
  }
}
```
- Functional component
```
function Test() {
  store.dispatch(); // Call to update state.
  return <div>Test</div>
}
```

All above dispatch calls will not be reflected with the new state in the `window.__PRELOADED_STATE__` if we prefetch state before `renderToString()`

According to the code snippet from [http://redux.js.org/docs/recipes/ServerRendering.html#handling-the-request](http://redux.js.org/docs/recipes/ServerRendering.html#handling-the-request), the state is grab after renderToString.
```
// Render the component to a string
const html = renderToString(
  <Provider store={store}>
    <App />
  </Provider>
)

// Grab the initial state from our Redux store
const preloadedState = store.getState()
```

Additionally, I have provided a unit test `"should have state preloaded after rendering"` to simulate the problem in this PR.

### DIFF
--- a/packages/electrode-redux-router-engine/lib/redux-router-engine.js
+++ b/packages/electrode-redux-router-engine/lib/redux-router-engine.js
@@ -117,17 +117,19 @@ class ReduxRouterEngine {
 
     return this._getReduxStoreInitializer(route, options).call(this, req, match)
       .then((store) => {
-        const r = { prefetch: stringifyPreloadedState(store.getState()) };
+        const r = {};
         const x = this._renderToString(req, store, match, withIds);
         if (x.then !== undefined) { // a Promise?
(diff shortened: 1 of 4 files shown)


jchip:
thanks for the fix.

Issue #474, opened 2017-07-13
Allow inline scripts to be marked with a nonce for CSP protection


bodawei:
If one is using A CSP HTTP header for security protection and wishes to lock down scripts, one may choose to mark any inline scripts (e.g. the one that holds the preloaded state) with a 'nonce'. That is, one would like:

```
<script nonce="<<<BASE64 VALUE>>>">
  window.__PRELOADED_STATE__ = ...;
</script>
```

One should be able to pass such a nonce into the react-webapp plugin and have that attached to any of the scripts that are included.

See also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src

bodawei:
I'm considering providing a PR to do this. My approach would be:

1) assume some plugin will generate the nonce and that this value will be available somewhere off of hapi's `request.plugins` (e.g. `request.plugins[pluginName][noncePropName]`)

2) Then, one can provide a routeOption like cspNonceProp which specifies where to find this off of the request (defaults to undefined).  Thus, I might configure `cspNonceProp='plugins.awesomeCSPPlugin.nonceValue`

3) Where a literal `<script></script>` is generated in `react-webapp.js`, if there is a cspNonceProp value set and that property chain can be found off of the `options.request` object, then add a `nonce="${extracted nonce value}"` to the script tags.

PR #486, opened 2017-07-19
[minor][feat] Allow inline scripts to be marked with a nonce for CSP protection

bodawei:
Modified react-webapp to notice when a nonce value has been provided, and if so attaches
it to any <script> tags and the critical css tag that are generated. See also https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

Fix for #474 

Note: Internal feedback suggested that in some cases a function would be a more general way to get the none value, so added this as an alternative.  Modified docs and tests to test both with a string and the function.


bodawei:
The continuus-integration tests appear to have failed in generator-electrode, which is unrelated to my change.  Am I reading this wrong?

jchip:
where are you adding the `Content-Security-Policy` HTTP header?

bodawei:
Adding it with a plugin called blankie.  This is just the change to catch the value from that, or any other similar plugin.

PR #507, opened 2017-07-30
[minor][feat] update component generator UI & using random name for component-add

didi0613:
Beautify the component generator UI, updated as a expanded accordions. This PR paired with electrode-demo-helper PR: https://github.com/electrode-io/electrode-demo-helper/pull/8

Tested in electrode:component and electrode:component-add generator.

Component Demo:
screen shot 2017-07-30 at 12 13 43 pm

Component-add Demo:
screen shot 2017-07-30 at 12 06 11 pm

### DIFF
--- a/packages/generator-electrode/component-add/index.js
+++ b/packages/generator-electrode/component-add/index.js
@@ -104,26 +104,13 @@ module.exports = generators.Base.extend({
     },
 
     askFor: function() {
-      function randomComponentName() {
-        var charSet = "abcdefghijklmnopqrstuvwxyz";
-        var randomString = "";
-
-        // Generate 7-digit random component name
-        for (var i = 0; i < 7; i++) {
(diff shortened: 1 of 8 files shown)


packages/generator-electrode/component-add/index.js
```diff
@@ -104,13 +104,26 @@ module.exports = generators.Base.extend({
     },
 
     askFor: function() {
+      function randomComponentName() {
```
jchip: it'd be a bit more consistent to use `untitled###` where `###` is a random number.

didi0613: Updated as: "untitled" + Math.floor(Math.random() * 1000)
Random name sample
```
? Component Name untitled228
? What is the ClassName for your component? 
? What will be the npm package name? untitled228
? What will be the GitHub organization username (e.g., 'walmartlabs')? 
```

PR #515, opened 2017-07-31
Getting ui-config and ui-logging to work with open-source sample app

dvonlehman:
I refactored the `samples/universal-react-node` app to make use of the latest versions of `electrode-react-webapp`, `electrode-react-context`, `electrode-ui-config`, `electrode-cookies`, and `electrode-ui-logger`. In the course of getting everything working I discovered that some changes needed to be made in the core modules. Most significantly `electrode-ui-config` was expecting a global variable named `window._wmt` to be defined. This is a holdover from when the code was Walmart specific. I have renamed the variable `window._app` (although @jchip may prefer something like `window._electrodeApp`). I also made an update to `electrode-react-webapp` to emit a `window._app.config.ui` to the `<head/>`. The value comes from `request.server.settings.app.config.ui`.

New versions of all the aforementioned modules will need to be published to npm. Since this is my last day working on Electrode, I leave it to @jchip to take this on.

In the `universal-react-node` sample I added a new demo page `app-context` which demonstrates how to use logging, config, and cookies from a React component - including passing the `context.app` appropriately.

One other thing. For some reason the build/tests are failing due to ES6 syntax in the `index.js` for the ISO modules (config, cookies, logging). It seems like webpack is pulling in the server version rather than just the module indicated by the `browser` property in package.json. Might be something to keep an eye out for in case it is leading to bundle bloat. I've gone ahead and removed all ES6 syntax.


jchip:
Thank you @dvonlehman.  This is awesome.  I will take your branch and update it.

PR #538, opened 2017-08-10
[minor][feat] implement code splitting support with webpack 3.

khayong:
Thanks for upgrading to webpack 3. With webpack 3, we could specify chunk name for each dynamically loaded JS file now.

```
import(/* webpackChunkName: "splitting" */ "./components/demo-code-splitting").then(module => {
  callback(null, module.default);
});
```
Beside that, it use another set of babel plugins (`babel-plugin-dynamic-import-webpack`, `babel-plugin-dynamic-import-node`) to implement.

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/babel/babelrc-client
+++ b/packages/electrode-archetype-react-app-dev/config/babel/babelrc-client
@@ -48,7 +48,8 @@
     ],
     "transform-node-env-inline",
     "lodash",
-    "transform-runtime"
+    "transform-runtime",
+    "syntax-dynamic-import"
   ],
   "env": {
     "production": {
--- a/packages/electrode-archetype-react-app-dev/package.json
+++ b/packages/electrode-archetype-react-app-dev/package.json
@@ -25,6 +25,8 @@
     "babel-core": "^6.4.0",
     "babel-eslint": "^7.0.0",
     "babel-loader": "^7.0.0",
+    "babel-plugin-dynamic-import-node": "^1.0.2",
+    "babel-plugin-dynamic-import-webpack": "^1.0.1",
     "babel-plugin-i18n-id-hashing": "^2.1.0",
     "babel-plugin-lodash": "^3.1.3",
     "babel-plugin-minify-dead-code-elimination": "^0.1.7",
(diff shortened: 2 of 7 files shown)


jchip:
thanks for doing this.  how does this might affect existing apps?

khayong:
On the file src/client/routes.jsx, instead of using JSX to describe routes, it should be described with plain route objects. 
```
export const routes = {
  path: "/",
  component: Home,
  childRoutes: [{
    path: "splitting",
    getComponent(nextState, callback) {
      import(/* webpackChunkName: "splitting" */ "./components/demo-code-splitting").then(module => {
        callback(null, module.default);
      });
    }
  }]
};
```

- Path `/` loads `Home` component statically into main js `main.bundle.dev.js`.
- Path `/splitting` dynamically load the component which is imported by `import()` syntax. When user navigate to the path, a new js file is loaded to browser with the file name [hash].splitting.js.

As `/splliting` is a child route of the root route `/`, the component of `/splitting` is placed as a child of `Home` component. To output the splliting component, we would need to call `{children}` under `Home` component.

Beside that, we need to use `Link` to load the splitting path as below. Once user click on the link, the content will be output as a child of the `Home` component.
```
import { Link } from "react-router";
render() {
  <Link to="/splitting">Load Here</Link>
}
```

jchip:
is the `/splitting` path a fixed path?  Meaning that all apps will need to use that?

khayong:
It's not necessary. The path is just a sample. 
You could define any child routes to be dynamic or not.

More sample can be found at below code.
[https://github.com/khayong/electrode-app/blob/mantra/src/client/modules/core/routes.jsx](url)

In case you are interested, the above sample repos is built by applying [Mantra Architecture](https://kadirahq.github.io/mantra/) to Electrode framework. Although the architecture is for Meteor application, we could make use of it for any frontend development. I modified it to use React Router + Redux + Code Splitting.

jchip:
sorry for the delay.  thanksgiving shopping coming.

would you mind documenting this please?  by adding a new section under Intermediate

jchip:
There was some issue with the CI and the changes in this PR was not tested.

The changes breaks the generated app when running unit tests.  I have reverted the changes.  With the fix from this commit https://github.com/electrode-io/electrode/commit/32c1598c3e0b76dfdb339227db02b34e9e62721a, could you please run `clap test-generator` at the top level to check the breakage please?  Thanks.

There's some minor eslint errors.  After that's fix, you'd see the breakage.

khayong:
I have rewritten the code, please review the PR at #610 . Thanks.

PR #549, opened 2017-08-16
[major] Update ESLint from v3 to v4

nucleartide:
Hello! 👋 

I work for Jet, and I'm running into some problems when using `electrode-archetype-react-component(-dev)`. Our shared ESLint config assumes ESLint v4, but when `electrode-archetype-react-component(-dev)` is installed, the ESLint version gets bumped down to v3. To illustrate:

```bash
$ ./node_modules/.bin/eslint -v
v3.19.0
```

Sadly, running ESLint v3 breaks some of our linting rules. So I've updated the ESLint version to v4.

Webpack also complains when I run `clap npm:prepublish` on the latest e-a-r-c version. Tweaking the `"entry"` in the Webpack config seems to fix it.

Do these fixes seem okay? Lemme know what you think!

### DIFF
--- a/packages/electrode-archetype-react-component-dev/config/webpack/partial/entry.js
+++ b/packages/electrode-archetype-react-component-dev/config/webpack/partial/entry.js
@@ -3,5 +3,5 @@
 const Path = require("path");
 
 module.exports = {
-  entry: Path.join(process.cwd(), "src")
+  entry: Path.join(process.cwd(), "src/index.js")
 };


packages/electrode-archetype-react-component-dev/config/webpack/partial/entry.js
```diff
@@ -3,5 +3,5 @@
 const Path = require("path");
 
 module.exports = {
-  entry: Path.join(process.cwd(), "src/index.js")
+  entry: Path.join(process.cwd(), "src")
```
cross311: Can we do these changes seperately?  Also do we know what this affects?

jchip:
Hi, thanks for the update. 
Any idea on what's the breaking changes of eslint 3 to 4?  I am ok with the update, but we have to understand what's the impact on existing components, and if we want to major bump the component archetype because of this.

Also, I agree with @cross311, please separate the entry change into another PR.  Thanks.

nucleartide:
@jchip I've split out the entry change into https://github.com/electrode-io/electrode/pull/551. 👍 

Also in favor of a major bump if that's okay with you, IMO any breaking changes upstream should bump the major version.

PR #556, opened 2017-08-31
[major] drop PropTypes for production

didi0613:
Same changes for wml archetype app: https://gecgithub01.walmart.com/electrode/electrode-archetype-react-app/pull/243/files

Aim at: Remove unnecessary React propTypes from the production build, as they are only used in development.


ajoshi0:
LGTM @didi0613

PR #561, opened 2017-09-08
Keep Status code from content and set status codes for Status 404

codingcircus:
# Logic for keeping Status Codes

## Introduction
The current ssr logic, provides no clear hook, how to set status codes headers. 

Let's say for example I have an application, that manages a book collection, with a single books route `/books/:name`. Let's say this route handles the display of the book with the given name, by fetching the book via redux. If the book is found, it displays Details and the http status 200, that is sent from the node server is ok. However if the book isn't found, currently we would still have an http status of 200 even though in the app we might display a 404 error message.

## Proposal
In the index-view.js of the application, we currently can prefetch data. Even now it wouldn't be too difficult to hook into the promise chain of the `app.routesEngine.render(req)` command at the end, and update the status code by checking the prefetched redux state. 

Consider this code example: 
```
const prefetchedStatePrefix = 'window.__PRELOADED_STATE__ = ';

const getPrefetchedState = state => JSON.parse(state.replace(prefetchedStatePrefix, '').replace(/;([^;]*)$/, '$1'));

const get404Response = res => ({
  ...res,
  status: 404,
});

module.exports = (req) => {
  const app = req.server && req.server.app || req.app;
  if (!app.routesEngine) {
    app.routesEngine = new ReduxRouterEngine({routes, createReduxStore});
  }

  // After the content was prefetched, check if book could be fetched, otherwise set 404 status
  return app.routesEngine.render(req)
    .then((res) => {
      const prefetchedState = getPrefetchedState(res.prefetch);
      if (!prefetchedState.books.wasNotFound) {
        return get404Response(res);
      }

      return res;
    });
};
```

Maybe in the future you could even provide a cleaner documented hook, to check for the content of the prefetch, but for now this would suffice. 

## Content of PR
I updated the content of the electrode-react-webapp package, to correctly use the status code of the prefetched ssr content, when sending the server response. 

## Tested

- hapi: tests are still running through without problems
- express: was tested manually in our application and behaves as expected

## Untested

- koa: I implemented this by looking at online documentation for koa and the current implementation. However I have no way of really testing this right now.

### DIFF
--- a/packages/electrode-react-webapp/lib/express/index.js
+++ b/packages/electrode-react-webapp/lib/express/index.js
@@ -2,12 +2,8 @@
 
 const _ = require("lodash");
 const assert = require("assert");
+const HttpStatus = require("http-status-codes");
 const ReactWebapp = require("../react-webapp");
-
-const HTTP_OK = 200;
-const HTTP_NOT_FOUND = 404;
-const HTTP_ERROR_500 = 500;
-const HTTP_REDIRECT = 302;
 
 const registerRoutes = (app, options, next) => {
   ReactWebapp.setupOptions(options)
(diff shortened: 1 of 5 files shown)


codingcircus:
# Update 09/11/2017

I implemented more status codes. 410 is handled now as well (might be important for SEO purposes) and error codes get better error messages.

For better readability I added the very small http-status-code (3kb) package.

codingcircus:
Unfortunately this doesn't work with the updated electrode-react-webapp (Version 2.0.0) anymore, so I will close this and investigate how to do it with the new version.

Issue #560, opened 2017-09-07
Custom 404


codingcircus:
Is it possible to send custom 404 pages with the Electrode Redux Router Engine, when some redux actions return an error?

Let's say for example I have an application, that serves books. When I navigate to a single book route (e.g. books/:name/) it would always return an https status of 200, even if the name doesn't exist. For SEO purposes it would be good, when when we could "hijack" the http status and still set it to 404.

codingcircus:
I wrote a pull request for exactly this issue. https://github.com/electrode-io/electrode/pull/561

jshado1:
I believe this is already possible by setting up middleware in `express-server.js` (which has access to the `response` object).

seansean11eis:
@jshado1 after upgrading to `electrode-react-webapp`, it is possible to intercept the response object (since the `status/code` bugs were fixed), but this would require modifying the response after it was sent by `electrode-react-webapp`.

I think the `electrode-react-webapp` and `electrode-redux-router-engine` libraries might need to be modified to allow us to set up custom status handlers.

At the moment, the [`handleRoute` function](https://github.com/electrode-io/electrode/blob/master/packages/electrode-react-webapp/lib/express/index.js#L22) reverts to `response.status(status).send(data);` when a request receives a 404 status from the `electrode-redux-router-engine`.  This will always return a json object to the browser.  As far as I know, there is no way to respond with a 404 HTML page to the browser.  However, the walmart.com website seems to handle them correctly, which makes me wonder if they are using the same library.

The potential issue in the `electrode-redux-router-engine` library is that if you set up a catch-all (`*`) route for client-side 404's in your `routes.jsx` file, it will always return a 200 response code for all routes.  The library checks to see if the route that the user hit has a match in `routes.jsx` and since `*` catches any 404, you will always be returned a 200 status for the `electrode-redux-router-engine`.  This may need to be modified, unless there is some way around the problem on the client side.

I'd like to start working on a PR, but would also like to hear opinions/thoughts if you have any.

codingcircus:
I already have an implementation for handling status codes in the express server in a forked version of version 2. So far I’d catch 404, 410 and 503 status code with custom html, as well as 301, 302, 307 and 308 with redirects. Do you think there is a need for other codes as well? 

Hopefully I can add it to Hapi and Koa as well in the next few days and open a new PR.

seansean11:
@codingcircus sorry didn't realize you were working on this.  Interested in seeing your solution as well.

[...]

PR #575, opened 2017-09-20
custom-404: allow for custom 404 responses

seansean11:
**DO NOT MERGE**

This is code to start the conversation around custom 404 handling in https://github.com/electrode-io/electrode/issues/560.  It still needs additional work and tests.  See the issue for details.

##  Context
For SEO reasons, we need to be returning a custom 404 response from the server on routes that do not exist.  Currently, if you set a catch-all route in `routes.jsx` you will always get a 200 response from the server on routes that don't exist.  If you do not have a catch-all route set up for React Router, the server will return a 404 response in the form of a JSON object.

## Solution
This PR would allow for a `notFoundRoute` property to be passed to the options object in `ReduxRouterEngine` render method in the `index-view.jsx`.  Passing this property allows the server side 404's to use the 404 not found route in `routes.jsx`.  The code for enabling this works similar to this [stack overflow post](https://stackoverflow.com/questions/35524117/express-status-404-with-react-router).  This allows the server to respond to the browser with the 404 route specified in the `routes.jsx` file.

screen shot 2017-09-21 at 8 05 46 am
screen shot 2017-09-21 at 8 05 25 am

### DIFF
--- a/packages/electrode-react-webapp/lib/express/index.js
+++ b/packages/electrode-react-webapp/lib/express/index.js
@@ -19,7 +19,7 @@ const handleRoute = (request, response, handler) => {
       } else if (status >= 200 && status < 300) {
         response.send(data.html !== undefined ? data.html : data);
       } else {
-        response.status(status).send(data);
+        response.status(status).send(data.html !== undefined ? data.html : data);
       }
     })
     .catch(err => response.status(err.status).send(err.message));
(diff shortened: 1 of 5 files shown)


jchip:
is this related to https://github.com/electrode-io/electrode/pull/580?

seansean11eis:
@jchip they are related, I've dropped a comment in #580 .

jchip:
should I look at this or 580 or both?

seansean11eis:
@jchip Can you take a look at both?  I think these two PR's both need some input and/or direction from someone like yourself who knows the packages well.  I think they could be combined in some way.  I have added another comment to https://github.com/electrode-io/electrode/pull/580.  My only ask is that we allow for 404's to be handled properly when a user visits a route that does not exist in `routes.jsx`.  Currently you will be returned a JSON object with a status of 404, or you will be returned your catch-all/not found route with a status of 200.  I think we should return the catch-all/not found route with a status of 404, which is what this PR accomplishes.

PR #580, opened 2017-09-25
Keep custom status codes

codingcircus:
# Logic for keeping Status Codes

## Introduction
The current ssr logic, provides no clear hook, how to set status codes headers. 

Let's say for example I have an application, that manages a book collection, with a single books route `/books/:name`. Let's say this route handles the display of the book with the given name, by fetching the book via redux. If the book is found, it displays Details and the http status 200, that is sent from the node server is ok. However if the book isn't found, currently we would still have an http status of 200 even though in the app we might display a 404 error message.

## Proposal
In the index-view.js of the application, we currently can prefetch data. Even now it wouldn't be too difficult to hook into the promise chain of the `app.routesEngine.render(req)` command at the end, and update the status code by checking the prefetched redux state. 

Consider this code example: 
```
const prefetchedStatePrefix = 'window.__PRELOADED_STATE__ = ';

const getPrefetchedState = state => JSON.parse(state.replace(prefetchedStatePrefix, '').replace(/;([^;]*)$/, '$1'));

const get404Response = res => ({
  ...res,
  status: 404,
});

module.exports = (req) => {
  const app = req.server && req.server.app || req.app;
  if (!app.routesEngine) {
    app.routesEngine = new ReduxRouterEngine({routes, createReduxStore});
  }

  // After the content was prefetched, check if book could be fetched, otherwise set 404 status
  return app.routesEngine.render(req)
    .then((res) => {
      const prefetchedState = getPrefetchedState(res.prefetch);
      if (!prefetchedState.books.wasNotFound) {
        return get404Response(res);
      }

      return res;
    });
};
```

Maybe in the future you could even provide a cleaner documented hook, to check for the content of the prefetch, but for now this would suffice. 

## Content of PR
I updated the content of the electrode-react-webapp package, to correctly use the status code of the prefetched ssr content, when sending the server response. 

## Handled Status Codes

for custom html I currently added the status codes 404, 410, 503. These are probably the most important status codes for SEO purposes, where we still want to have a custom html output instead of json messages. 

for redirects I also added 301, 307 and 308 to all server implementations, however I only set specific statuses when redirecting in the express implementation. 

## Tested

- hapi: tests are still running through without problems
- express: was tested manually in our application and behaves as expected. Additional tests were added and the old tests are still passing
- koa: tests are still running through without problems


jchip:
Is this something that can be done with react-helmet?  We had react-helmet built-in with react-webap 1.0, but I've removed that in 2.0 and demo using it as an add-on in the tests.

codingcircus:
We didn't find any way to do it via helmet. I think the main problem is, that the status needs to be already set, when you get the response from the node server. Otherwise the ssr will have a different status code, then the client render.

seansean11:
@codingcircus if you have any feedback for my PR let me know https://github.com/electrode-io/electrode/pull/575.  I figured I would weigh in here because our PR's have some common ground (mainly handling 404 responses).

There are a couple of questions I have about the PR because I first tried to handle 404's from within the index-view, in the `then` block of the Redux Router Engine `render` method (I wasn't successful).  Correct me if I'm wrong, buy I am fairly certain that this method only returns an object like so `{ status, html, prefetch }`, which would mean that there is no access to the `res` object at this point in the code.

If that express `res` object were to be available in the `then` block, I might also have some reservations about returning a status code such as `404` based on prefetched state.  Ideally, I think we would want a 404 HTML page returned from the server when a user hits a route that does not exist in the `routes.jsx` file, or hits your 'not found' catch-all route in `routes.jsx`.  

One way of accomplishing this, would be to make sure that `app.routesEngine.render(req)` correctly returns an object with a `status` of 404, and also make sure that this is correctly handled by electrode-react-webapp.

Since the Redux Router Engine is already set up to return objects with a status code, I thought it might be easiest to utilize this (whether 404 status or otherwise) and then handle these codes appropriately via the electrode-react-webapp.  Although, if the library were re-architected in a way that we could tap into custom status handling this may be quite a bit easier.  Let me know what you think.

codingcircus:
@seansean11 As far as I can see, you have a smiliar solution, however you don't handle all the status codes, that we will need in our application. Also what was not working for us so far was having a custom html with a 404, because as soon as the react-webapp encounters an object with an html key value pair, it defaults to a 200 status code (see my change in `render-output.js`). 

The reason why we're using the prefetched state for setting the status code in our application is, that we have routes where it's not immediately clear, what the status is, since we have asynchronous data fetches, that need to be resolved, before we know more. I think the cleanest solution for this is precisely using the prefetched state, since it knows, if the fetch returned a result or a not found. 

Hope this clarifies a little bit, what we try to achieve.

packages/electrode-react-webapp/lib/express/index.js
```diff
@@ -5,17 +5,35 @@
...
+        HttpStatus.MOVED_TEMPORARILY,
+        HttpStatus.PERMANENT_REDIRECT,
+        HttpStatus.TEMPORARY_REDIRECT
+      ];
+
       if (status === undefined) {
         response.send(data);
-      } else if (status === HTTP_REDIRECT) {
-        response.redirect(data.path);
+      } else if (redirectStatuses.find(redirectStatus => redirectStatus === status)) {
```
jchip: why not use `indexOf`?

seansean11eis:
@codingcircus I agree that it's nice to have the options to handle the different status codes.  How would you send a `res` with your specifically designed 404 page with these changes?  Are you pulling in a custom HTML file in `index-view.jsx` and assigning its HTML to the html prop in the object returned from `ReactRouterEngine.render`?  The reason I ask is that the express.js `res` object is not available to `res.send()` in this file, so I would imagine you would need to send the HTML to the handler in `electrode-react-webapp`.

In my case, I simply want to return a 404 page with a status of 404 when the user visits a route in the UI that doesn't exist in routes.jsx.  The fact that this doesn't exist at the moment seems like a bug to me.  It currently returns a JSON object to the browser with a status of 404, or a 200 response if you have a "not found" or catch-all route in `routes.jsx`.  Since `ReactRouterEngine.render` already responds with 500, 404, and 302 objects, my PR leverages this to return a 404 route from the `routes.jsx` file, with a status of 404.  This way the 404 page still gets rendered via ReactRouterEngine.  Our PR's share a couple of the same fixes, but achieve different things.

If a user visits an existing route in the UI, I would think that they should get a 200 response. If the route exists, but the prefetched state is not available for that route, I wouldn't necessarily want to send a 404 status response from the server to the user.  With that said, having this available may be helpful in certain use cases.

Pending feedback from the Electrode team, maybe we can find a way to combine our 2 concerns.

packages/electrode-react-webapp/lib/express/index.js
codingcircus: Personal preference 😉 If you want, I can change this.

codingcircus:
Well the current way how we handle this in our fork is, we have a separate reducer state for status codes and we have a component, that is basically the content of our 404 page, that sets the reducer state in componentWillMount to 404. Since this is not an asynchronous dispatch, it doesn't even need to be added to index-view.js, the only thing we do there is check the result of the render for the status-code state.

```
return app.routesEngine.render(req)
    .then(result => updateStatusCode(result));

export function updateStatusCode (result) {
  const prefetchedState = getPrefetchedState(result.prefetch);

  // Check the status code reducer for current http code
  const {statusCodes} = prefetchedState;
  switch (statusCodes.code) {
    case HttpStatus.NOT_FOUND:
      return getResponseWithCustomStatus(result, HttpStatus.NOT_FOUND);
    case HttpStatus.MOVED_PERMANENTLY:
    case HttpStatus.MOVED_TEMPORARILY:
      return {
        status: HttpStatus.MOVED_PERMANENTLY,
        path: prefetchedState.statusCodes.meta.redirectPath,
      };
    default:
      return result;
  }
}
```

However I disagree, that all specified routes should only be handled as 200, since, as I explained in my PR, it might be that the route can be resolved, however the content of the route returns a 404 from the Database.

seansean11eis:
Thanks @codingcircus, I definitely understand now and I think it adds some nice flexibility to the framework.  I think our PR's should be looked at separately by the Electrode team because they are solving separate issues.

esimons:
@seansean11 we also need to support 404s based on prefetched data, and have a feature coming up to support 301 redirects (also based on prefetched data). So I think it's going to be really important for our work to be able to attach a status code to a technically-valid route as is proposed here.
I appreciate the use-case that you've brought up, though, which really does seem distinct from this. I forget what the `renderProps` look like that react-router returns from [a `match` call](https://github.com/ReactTraining/react-router/blob/v3/docs/API.md#match-routes-location-history-options--cb), but wonder if it'd be possible to check that object for the presence of some prop that we determine to be our convention, and if a matching route contains it to set the returned status code to 404? So declare a route like:
`<Route path="*" component={NotFound} server404 />`
And then [in the redux-router-engine code](https://github.com/electrode-io/electrode/pull/575/files#diff-86fb219987335569ce41401851671a43R74) check for that `server404` prop on matched components?

seansean11eis:
@esimons 👍  that was also one of the ideas I was also thinking of at the start of https://github.com/electrode-io/electrode/pull/575.  The only reason I didn't go down that route (no pun intended) was because it would mean that Electrode needs to assume a specific convention for the `routes.jsx` file.  I like this idea as well though, and wouldn't mind either approach.

It sounds like there is a definite need for this PR though.

jchip:
Hi, I managed to mess up the commits here.
While testing the changes locally, I found that `http-status-codes` was only added to `devDependencies`.

I've taken your changes and rebase with the latest master.

I also reformat the new code with prettier and refactor the http status code check slightly.

There are a few lines of new code that's not covered by unit test.

The branch is here: https://github.com/jchip/electrode/commits/keep-custom-status-codes

could you please check the coverage?  Coverage check is done by `clap check`, but I've updated `npm test` to run that.

Thanks.  We can close this PR and open a new one.

devourment77:
Hello,  I know this is old, but is this functionality covered in the electrode docs?

Issue #102, opened 2017-01-05
Confused how to run server tests in watch mode


citelao:
I've tried `gulp server-test` and `gulp test-watch-all` but I can't for the life of me figure out how to setup auto-running server-side tests.

I can run tests manually with `gulp test-dev`, but the other two just start up a file browser at `localhost:3001`. 

How do I do this?

jchip:
`test-watch` is suppose to do this.  It's very possible that is broken at the moment since it's not part of the CI verification.  Thanks for reporting this.

DavidOrchard:
test-watch should be fixed in next release.

sergesemashko:
I have the latest app from electrode generator and it gives me `Chrome 60.0.3112 (Mac OS X 10.11.4): Executed 0 of 0 ERROR (0.041 secs / 0 secs)` in `test-watch`, however `clap check` works fine.  
Any ideas how to solve that?

sergesemashko:
It turned out that karma launched browser cannot get webpack builld due to CORS issue. Following code in my custom webpack dev config fixed the issue: 
```js
////archetype/config/webpack/webpack.config.hot.js

module.exports = function( composer, options, compose ) {
  // Now that the composer has been updated, compose the final config
  const config = compose();
  config.devServer.disableHostCheck = true;
  config.devServer.headers = {
    "Access-Control-Allow-Origin": "*"
  };
  return config;
};

```

PR #584, opened 2017-09-27
[patch][bug] Fix CORS issue when running karma tests in `test-watch` and `test-watch-all`

sergesemashko:
This PR fixes the [CORS issue](https://github.com/electrode-io/electrode/issues/102#issuecomment-332284482) when karma is running in `test-watch` and `test-watch-all` modes.

As the result of issue none of the test is running in watch mode. Mocha output:
`Chrome 60.0.3112 (Mac OS X 10.11.4): Executed 0 of 0 ERROR (0.041 secs / 0 secs)`.

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/test-base.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/test-base.js
@@ -14,7 +14,17 @@ module.exports = {
     noParse: [new RegExp(Path.normalize("node_modules/sinon/").replace(/\\/g, "\\\\"))]
   },
   devServer: {
-    stats: "errors-only" // only show errors
+    stats: "errors-only", // only show errors
+    /*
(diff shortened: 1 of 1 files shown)


packages/electrode-archetype-react-app-dev/config/webpack/partial/test-base.js
```diff
@@ -14,7 +14,17 @@ module.exports = {
...
   devServer: {
-    stats: "errors-only" // only show errors
+    stats: "errors-only", // only show errors
+    /*
+     * Karma browsers and webpack are running on different ports, which introduces
+     * CORS issue.
+     * The following code allows karma launched browsers to access webpack
+     * bundle in `test-watch` and `test-watch-all` modes.
+     */
+    disableHostCheck: true,
```
sergesemashko: I slightly doubt that this is the right place for this configuration. If there is a better idea - let me know

PR #610, opened 2017-10-12
[minor][feat] implement dynamic code splitting

khayong:
Incorporated babel plugin [babel-plugin-dynamic-import-node](https://github.com/airbnb/babel-plugin-dynamic-import-node) to support [dynamic code splitting](https://webpack.js.org/guides/code-splitting/#dynamic-imports).

Added a sample code of lazy loading practice, the example was taken from [Lazy loading Guide](https://webpack.js.org/guides/lazy-loading/).

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/babel/babelrc-client
+++ b/packages/electrode-archetype-react-app-dev/config/babel/babelrc-client
@@ -48,7 +48,8 @@
     ],
     "transform-node-env-inline",
     "lodash",
-    "transform-runtime"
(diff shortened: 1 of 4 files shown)


cristoni:
Having this feature in the master branch would be really nice.

jchip:
Hi, thanks for the PR, we can't merge this because the original PR has samples on using this, but it failed.  The update to pass the failure was to just remove the samples.  We don't know what are the regress possibility of merging this.  We will look into implementing this with working samples.

PR #612, opened 2017-10-13
Reports code coverage on server side tests

mcrowder65:
Assuming the user is using `"istanbul": "^1.1.0-alpha.1"`, this generates a code coverage report like so, as well as lcov.info file.. We can change this archetype's istanbul download to be the alpha if you want. right now code coverage doesn't work anyways... we explicitly installed istanbul in our root instead of using 0.4.5 and that's what made it work
image


ajoshi0:
LGTM @mcrowder65

mcrowder65:
Huh

mcrowder65:
@ajoshi0 hahaha I looked up what LTGM stood for and it said “learn the game mechanics”, meaning something like you suck and have no idea what you’re doing so I was ready be like what the heck man!!! Then I looked up LGTM :)

ajoshi0:
LGTM --> Looks good to me @mcrowder65 :)

PR #615, opened 2017-10-17
[minor] Changes for wml electrode-ignite

didi0613:
- yopath for generators accept params for base
- gitbook accept params for wml link
- update unit test
- fixed a hanging case when child process exit its own for any reason

### DIFF
--- a/packages/electrode-ignite/lib/menu-items/electrode-oss-doc.js
+++ b/packages/electrode-ignite/lib/menu-items/electrode-oss-doc.js
@@ -5,12 +5,11 @@ const Promise = require("bluebird");
 const opn = require("opn");
 const chalk = require("chalk");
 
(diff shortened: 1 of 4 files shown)


packages/electrode-ignite/lib/menu-items/electrode-oss-doc.js
```diff
@@ -5,9 +5,10 @@ const Promise = require("bluebird");
 const opn = require("opn");
 const chalk = require("chalk");
 
-module.exports = function() {
+module.exports = function(link) {
```
jchip: should add a `docs` so the `logger.log` below doesn't always log the same `"Electrode docs"`.

didi0613: done @jchip

PR #620, opened 2017-10-19
[patch][bug] Component Generator refactor install function, exit immediately.

animesh10:
- npm/yarn install exits immediately on interrupt.

### DIFF
--- a/packages/generator-electrode/component/index.js
+++ b/packages/generator-electrode/component/index.js
@@ -455,9 +455,9 @@ module.exports = class extends Generator {
       ];
     }
 
-    let res = locations.reduce(
-      (previousValue, currentValue, currentIndex, array) => {
-        if (previousValue.signal != "SIGINT" && previousValue.status === 0) {
+    locations.reduce(
+      (previousValue, currentValue) => {
+        if (!previousValue.signal && previousValue.status === 0) {
           return this.spawnCommandSync(yarnOrNpm, ["install"], {
             cwd: currentValue
           });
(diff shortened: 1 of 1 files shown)


packages/generator-electrode/component/index.js
```diff
@@ -437,57 +437,36 @@ module.exports = class extends Generator {
...
-      this.spawnCommandSync("yarn", ["install"], {
-        cwd: this.destinationPath()
-      });
-    } else {
-      this.spawnCommandSync("npm", ["install"], {
-        cwd: this.destinationPath()
-      });
-    }
+    let res = locations.reduce(
+      (previousValue, currentValue, currentIndex, array) => {
```
jchip: `currentIndex` and `array` not used and are not needed.

packages/generator-electrode/component/index.js
```diff
@@ -437,57 +437,36 @@ module.exports = class extends Generator {
...
+        if (previousValue.signal != "SIGINT" && previousValue.status === 0) {
```
jchip: should just check `!previousValue.signal`

packages/generator-electrode/component/index.js
```diff
@@ -437,57 +437,36 @@ module.exports = class extends Generator {
...
-    if (this.props.yarn) {
-      this.spawnCommandSync("yarn", ["install"], {
-        cwd: this.destinationPath()
-      });
-    } else {
-      this.spawnCommandSync("npm", ["install"], {
-        cwd: this.destinationPath()
-      });
-    }
+    let res = locations.reduce(
```
jchip: `res` not used, why need it?

PR #628, opened 2017-11-02
Default reporter socket port from env

codingcircus:
# Override standard reporter socket port

Since we are currently running multiple electrode apps in our dev environment, we run into a conflict, since the Webpack Reporter always uses the same standard port. This PR introduces a new environment variable called WEBPACK_REPORTER_SOCKET_PORT, that can be used to override the standard port 5000.

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/archetype.js
+++ b/packages/electrode-archetype-react-app-dev/config/archetype.js
@@ -14,6 +14,7 @@ const webpackConfigSpec = {
   devHostname: { env: ["WEBPACK_HOST", "WEBPACK_DEV_HOST"], default: "localhost" },
   devPort: { env: "WEBPACK_DEV_PORT", default: 2992 },
   testPort: { env: "WEBPACK_TEST_PORT", default: 3001 },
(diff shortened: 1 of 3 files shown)


jchip:
I prefer this to be passed in to webpack reporter from the archetype.  And have it processed here https://github.com/electrode-io/electrode/blob/master/packages/electrode-archetype-react-app-dev/config/archetype.js#L16, and pass to WebpackReporter here https://github.com/electrode-io/electrode/blob/master/packages/electrode-archetype-react-app-dev/config/webpack/partial/html-reporter.js#L8.

codingcircus:
So if I understand correctly, we should just overwrite the default clap task `html-reporter`, is that right?

We already managed to do this, however, then I would at least propose, that you extend the current documentation, so that other developers find this information more quickly. 

Also my PR can be closed then.

jchip:
I meant it's OK to have an env variable, but I want the archetype to process that env variable instead.

codingcircus:
I updated my PR based on your last comments now

jchip:
sorry, not sure what the CI failure was about, but merging it and will check.

PR #635, opened 2017-11-15
[major] move to react 16 exclusively

jchip:


mcrowder65:
Doesn’t the generator use an older version of react router that does React.createClass which is now deprecated in 16?

didi0613:
lgtm

jchip:
@mcrowder65 react router 4 made some major changes and we need to figure out how to upgrade to it first.

PR #636, opened 2017-11-15
[patch] enable defer based on latest css-split-webpack-plugin

didi0613:
enable defer for deferring css split after optimization.


ajoshi0:
LGTM @didi0613

PR #637, opened 2017-11-15
[patch] more info in log output

jchip:


packages/electrode-archetype-opt-inferno/optional-check.js
```diff
@@ -13,29 +13,33 @@ function findAppDir() {
 }
 
 const appDir = findAppDir();
+const myPkg = require("./package.json");
```
datvong-wm: assumes the current directory has a package.json.

jchip: not assumes, ensure.

datvong-wm: then maybe test the cwd to ensure it's package root so we can display a descriptive message.

this is going to say "can't read package.json".. which isn't very informative.

PR #642, opened 2017-11-19
[patch] Few fixes for new electrode component

didi0613:
Fixes for new electrode component:
- Fix the issue when updating the package name in `package/[componentName]/package.json`, webpack hot reload will not work.
- update `eslint-config-walmart` to 2.0.0 for `jsx-wrap-multilines` rule (Need a new patch release for https://github.com/walmartlabs/eslint-config-walmart)
- minor fixes for UI (currently the accordion is too wide)
- fixes for component `clap test` eslint errors

### DIFF
--- a/packages/generator-electrode/component/templates/demo-app/archetype/config/webpack/component-alias.config.js
+++ b/packages/generator-electrode/component/templates/demo-app/archetype/config/webpack/component-alias.config.js
@@ -15,10 +15,11 @@ const config = {
   }
 };
 
-components.forEach(name => {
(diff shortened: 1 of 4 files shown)


packages/generator-electrode/component/templates/demo-app/archetype/config/webpack/component-alias.config.js
```diff
@@ -16,9 +16,10 @@ const config = {
 };
 
 components.forEach(name => {
```
psolidgold: This should be `folderName` correct?

didi0613:
@psolidgold yes right! thanks for pointing it out.

PR #652, opened 2017-11-29
Updated eslint-config-walmart to use the latest version

mcrowder65:


mcrowder65:
I’ll be coming back to this later when I have more time

PR #666, opened 2018-01-04
[major] upgrade electrode-redux-router-engine to react16

didi0613:


jchip:
this should not be needed.  it can work with React 14, 15, or 16.

PR #672, opened 2018-01-11
Fix broken Electrify chart due to stats.chunks data missing

aakula-edmunds:
We ran into issue with broken Electrify chart. And it turned out that `stats.chunks` was missing. If required stats info fields are specified explicitly `stats.chunks` is present and Electrify chart works fine.


aakula-edmunds:
I see CI build failed, but the failure was not caused my changes.

jchip:
Hi, thanks for the PR.  We are updating all of our modules to React 16.  It seems electrify have some conflicts that we haven't had time to investigate yet so we are temporarily removing it to unblock the React 16 upgrade for now.  We will revisit this once we have a stable upgrade to React 16.

jchip:
Hi, sorry, no one is maintaining electrode-electrify now and we don't have any ETA on getting it up to date with React 16.  It's removed so closing this for now.  Thanks for the PR.

PR #673, opened 2018-01-12
[major] update electrode app generator to support react16

didi0613:


packages/generator-electrode/generators/app/templates/_package.json
```diff
@@ -31,9 +31,9 @@
   },
   "dependencies": {
     "bluebird": "^3.4.6",
-    "electrode-archetype-react-app": "^4.0.0",
+    "electrode-archetype-react-app": "^5.0.0",
```
didi0613: Pending on the releases of "electrode-archetype-react-app": "^5.0.0" and "electrode-react-webapp": "^3.0.0".

jchip:
this will be taken care of by lerna when publishing

jchip:
actually, no, not related to lerna.

PR #703, opened 2018-01-26
[minor][archetype app] Update archetype app styles

didi0613:
Update archetype app styles
```
Use Cases
/*
 * cssModuleSupport: false
 * case 1: *only* *.css => normal CSS
 * case 2: *only* *.styl exists => Stylus
 * case 3: *only* *.scss exists => SASS
 * 
 * cssModulesSupport: true
 * case 1: *only* *.css => CSS-Modules + CSS-Next
 * case 2: *only* *.styl => normal CSS => CSS-Modules + CSS-Next
 * case 3: *only* *.scss => normal CSS => CSS-Modules + CSS-Next
 */
```

`cssModulesSupport true`: css-modules enabled, class names and animation names are scoped locally by default.

```
css => css-modules + css-next
stylus and scss => normal css first => css-modules +css-next
```
screen shot 2018-01-30 at 9 32 20 am

`cssModulesSupport: false` loaded as normal styles (css, stylus and scss)
false

`cssModuleStylusSupport ` is about to deprecate. Instead, please use "*.styl" + `cssModuleSupport: true`.

Verified on local sample app and home app.


didi0613:
About default value:

Original extract-styles default scenarios:
```
*.css* only:  css-modules + css next
*.styl* only + (default) cssModuleStylusSupport: false: normal css loader + stylus loader
*.styl* only | *.css+*styl, cssModuleStylusSupport: true: css-modules + stylus loader
```

```
Default value:
/*
 * cssModuleSupport initial value is undefined
 * if *only* *.css
 * cssModuleSupport default value is true
 * else ex) *only* *.styl|*.scss
 * cssModuleSupport default value is false
*/
```

PR #711, opened 2018-02-02
[minor] Update electrode-demo-index to v2 in component archetype

didi0613:


didi0613:
Keep the `electride-demo-index` in the archetype.
Updates for the version will be affected in the sample component and generators.

PR #715, opened 2018-02-05
Support webpack chunks (css)

### LINKED ISSUE #708: Electrode React Webapp does not read chunk names from webpack & arbitrarily appends to filenames
First, WHY would you [arbitrarily append hard-coded values of `'.style'` and `'.bundle.dev'`](https://github.com/electrode-io/electrode/blob/master/packages/electrode-react-webapp/lib/react-webapp.js#L107-L112) to file names set from variables!?

Second, instead of reading from the source of truth that is webpack, the [jsdoc for `util.resolveChunkSelector()`](https://github.com/electrode-io/electrode/blob/master/packages/electrode-react-webapp/lib/utils.js#L11) says it is based on some arbitrary configuration object, which I infer is [this](https://github.com/electrode-io/electrode/tree/master/packages/electrode-react-webapp#registering-with-electrode-server) (I say "infer" because I can't find anything that explicitly links them). This also causes asset multi-chunking (ex css) to break in Electrode React Webapp (which expects a string), but Webpack itself allows an array.

jshado1:
The app using chunks will need to implement the following to use this.

### Dev
```js
// /config/utils/chunk-selector.js
module.exports = () => ({
    css: [
        'app',
        'third-party-something',
    ],
    js: 'app',
});
```

```js
// /config/default.js

module.exports = {
    // …
    webapp: {
        // …
            options: {
                bundleChunkSelector: path.resolve(__dirname, './utils/chunk-selector'),
```

### Prod

```js
// /archetype/config/webpack/webpack.config.base.js
const appOverrides = {
    // …
    entry: {
        app: [
            'babel-polyfill',
            // ^ electrode-archetype-react-app-dev/config/webpack/partial/entry.js#L37
            path.resolve(appSrc.client, 'app.jsx')
        ],
        'third-party-something': path.resolve(appSrc.client, 'path/to/file'),
    },
};

const customMerger = (a, b) => {
    return (Array.isArray(a) && Array.isArray(b))
        ? [].concat(a).concat(b)
        : undefined;
};

module.exports = function webpackConfigBuilder(envOverrides = {}) {
    return function overrideElectrodeWebapp(composer, options, compose) {
        // this might be overkill (could probably just return the merge?)

        // …

        const config = compose();

        _.mergeWith(
            config,
            appOverrides,
            envOverrides,
            customMerger
        );
    };
};
```

```js
// /archetype/config/webpack/webpack.config.dev.js

const baseWebPack = require('./webpack.config.base');

module.exports = baseWebPack({
    devtool: '#cheap-module-eval-source-map'
});
```

---

I verified this against my own app using semantic-ui-less and react-semantic-ui. The following scenarios worked:
* Dev mode
  * a single value for `css` in chunkSelector
  * an array of 2 values for `css` in chunkSelector
* Prod mode
  * a single key-value pair within `entry` (webpack.config.base)
  * a second key-value pair within `entry` (webpack.config.base

I couldn't hook this up directly to webpack would creating a breaking change with implementers who are relying on the man-in-the-middle. So I think this is the next best and unblocks more functionality already available from webpack.

resolves #708

### DIFF
--- a/packages/electrode-react-webapp/lib/default-handlers.js
+++ b/packages/electrode-react-webapp/lib/default-handlers.js
@@ -62,11 +62,16 @@ module.exports = function setup(options) {
       const data = context.$.data;
       const manifest = bundleManifest();
       const manifestLink = manifest ? `<link rel="manifest" href="${manifest}" />\n` : "";
-      const css = WEBPACK_DEV
-        ? (!Array.isArray(data.devCSSBundle) && [data.devCSSBundle]) || data.devCSSBundle
-        : data.cssChunk
-          ? (!Array.isArray(data.cssChunk) && [data.cssChunk]) || data.cssChunk
-          : [""];
+      const css = [];
+
+      if (WEBPACK_DEV) {
+        if (Array.isArray(data.devCSSBundle)) css.push(...data.devCSSBundle);
+        else if (data.devCSSBundle) css.push(data.devCSSBundle);
+      }
+      else if (Array.isArray(data.cssChunk)) {
+        css.push(...data.cssChunk);
+      }
+      else if (data.cssChunk) css.push(data.cssChunk);
 
       const cssLink = css.reduce((acc, file) => {
         acc += `<link rel="stylesheet" href="${
--- a/packages/electrode-react-webapp/lib/default-handlers.js
+++ b/packages/electrode-react-webapp/lib/default-handlers.js
@@ -62,14 +62,11 @@ module.exports = function setup(options) {
       const data = context.$.data;
       const manifest = bundleManifest();
       const manifestLink = manifest ? `<link rel="manifest" href="${manifest}" />\n` : "";
-      const css = [];
+      let css = [];
 
-      if (WEBPACK_DEV) {
-        if (Array.isArray(data.devCSSBundle)) css.push(...data.devCSSBundle);
-        else if (data.devCSSBundle) css.push(data.devCSSBundle);
-      } else if (Array.isArray(data.cssChunk)) {
-        css.push(...data.cssChunk);
-      } else if (data.cssChunk) css.push(data.cssChunk);
+      if (data.criticalCSS); // leave css as empty array (reduce will turn it into an empty string)
+      else if (WEBPACK_DEV) css = css.concat(data.devCSSBundle);
+      else css = css.concat(data.cssChunk);
 
       const cssLink = css.reduce((acc, file) => {
         acc += `<link rel="stylesheet" href="${
--- a/packages/electrode-react-webapp/lib/default-handlers.js
+++ b/packages/electrode-react-webapp/lib/default-handlers.js
@@ -67,11 +67,9 @@ module.exports = function setup(options) {
       if (WEBPACK_DEV) {
         if (Array.isArray(data.devCSSBundle)) css.push(...data.devCSSBundle);
         else if (data.devCSSBundle) css.push(data.devCSSBundle);
-      }
-      else if (Array.isArray(data.cssChunk)) {
+      } else if (Array.isArray(data.cssChunk)) {
         css.push(...data.cssChunk);
-      }
-      else if (data.cssChunk) css.push(data.cssChunk);
+      } else if (data.cssChunk) css.push(data.cssChunk);
 
       const cssLink = css.reduce((acc, file) => {
         acc += `<link rel="stylesheet" href="${
--- a/packages/electrode-react-webapp/lib/default-handlers.js
+++ b/packages/electrode-react-webapp/lib/default-handlers.js
@@ -62,11 +62,9 @@ module.exports = function setup(options) {
       const data = context.$.data;
       const manifest = bundleManifest();
       const manifestLink = manifest ? `<link rel="manifest" href="${manifest}" />\n` : "";
-      let css = [];
-
-      if (data.criticalCSS); // leave css as empty array (reduce will turn it into an empty string)
-      else if (WEBPACK_DEV) css = css.concat(data.devCSSBundle);
-      else css = css.concat(data.cssChunk);
+      const css = WEBPACK_DEV
+        ? Array.prototype.concat(data.devCSSBundle)
+        : Array.prototype.concat(data.cssChunk);
 
       const cssLink = css.reduce((acc, file) => {
         acc += `<link rel="stylesheet" href="${
--- a/packages/electrode-react-webapp/lib/default-handlers.js
+++ b/packages/electrode-react-webapp/lib/default-handlers.js
@@ -63,9 +63,19 @@ module.exports = function setup(options) {
       const manifest = bundleManifest();
       const manifestLink = manifest ? `<link rel="manifest" href="${manifest}" />\n` : "";
       const css = WEBPACK_DEV
-        ? data.devCSSBundle
-        : (data.cssChunk && `${prodBundleBase}${data.cssChunk.name}`) || "";
-      const cssLink = css && !data.criticalCSS ? `<link rel="stylesheet" href="${css}" />` : "";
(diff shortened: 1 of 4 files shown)


packages/electrode-react-webapp/lib/default-handlers.js
```diff
@@ -63,9 +63,19 @@ module.exports = function setup(options) {
       const manifest = bundleManifest();
       const manifestLink = manifest ? `<link rel="manifest" href="${manifest}" />\n` : "";
       const css = WEBPACK_DEV
-        ? data.devCSSBundle
-        : (data.cssChunk && `${prodBundleBase}${data.cssChunk.name}`) || "";
-      const cssLink = css && !data.criticalCSS ? `<link rel="stylesheet" href="${css}" />` : "";
+        ? (!Array.isArray(data.devCSSBundle) && [data.devCSSBundle]) || data.devCSSBundle
+        : data.cssChunk
+          ? (!Array.isArray(data.cssChunk) && [data.cssChunk]) || data.cssChunk
+          : [""];
```
jshado1: This nested ternary isn't pretty, but it seems like you strong prefer `const`, and without, it would mean a `let`.

packages/electrode-react-webapp/lib/react-webapp.js
```diff
@@ -104,14 +104,31 @@ function makeRouteHandler(routeOptions, userContent) {
...
 
       const chunkNames = chunkSelector(request);
-      const devCSSBundle = chunkNames.css
-        ? `${devBundleBase}${chunkNames.css}.style.css`
-        : `${devBundleBase}style.css`;
-      const devJSBundle = chunkNames.js
-        ? `${devBundleBase}${chunkNames.js}.bundle.dev.js`
-        : `${devBundleBase}bundle.dev.js`;
-      const jsChunk = _.find(assets.js, asset => _.includes(asset.chunkNames, chunkNames.js));
-      const cssChunk = _.find(assets.css, asset => _.includes(asset.chunkNames, chunkNames.css));
```
jshado1: I'm not sure what the purpose of this was except to silently stomp file name changes that would have otherwise worked.

jshado1:
I dunno what's happening with hapi (I've never used it). When I run it in an app using express, the expected output from the server-agnostic utils, I get exactly the output that the assertion is trying to make.

jshado1:
Okay, I have no clue what you're trying to do in [electrode-react-webapp group-scripts.js](https://github.com/electrode-io/electrode/blob/master/packages/electrode-react-webapp/lib/group-scripts.js#L21) (the code is way too obfuscated and there are no comments). I'm gonna back out the change to the JS stuff (which is breaking hapi specs).

jshado1:
@jchip are you around?

jchip:
Thanks for the PR.  I am a bit busy with internal stuff at the moment.  I will get to this later today.  

group-scripts is trying to combine multiple consecutive literal JS code chunks into a single string, to avoid having a `<script>` tag for each.

packages/electrode-react-webapp/lib/default-handlers.js
```diff
@@ -62,10 +62,23 @@ module.exports = function setup(options) {
...
-        : (data.cssChunk && `${prodBundleBase}${data.cssChunk.name}`) || "";
-      const cssLink = css && !data.criticalCSS ? `<link rel="stylesheet" href="${css}" />` : "";
+      const css = [];
+
+      if (WEBPACK_DEV) {
+        if (Array.isArray(data.devCSSBundle)) css.push(...data.devCSSBundle);
+        else if (data.devCSSBundle) css.push(data.devCSSBundle);
+      } else if (Array.isArray(data.cssChunk)) {
+        css.push(...data.cssChunk);
+      } else if (data.cssChunk) css.push(data.cssChunk);
```
jchip: rather than `push`, `concat` would work better here and you don't have to do `isArray` check.

packages/electrode-react-webapp/lib/default-handlers.js
```diff
@@ -62,10 +62,23 @@ module.exports = function setup(options) {
...
+
+      const cssLink = css.reduce((acc, file) => {
```
jchip: the original code depends on the expression `!data.criticalCSS`.  Is there something I am missing that you don't need it here anymore?

packages/electrode-react-webapp/lib/react-webapp.js
```diff
@@ -104,14 +104,23 @@ function makeRouteHandler(routeOptions, userContent) {
...
+          cssChunks,
+          (chunkName) => `${devBundleBase}${chunkName}.style.css`
+        );
+      } else devCSSBundle = [`${devBundleBase}style.css`];
+
       const devJSBundle = chunkNames.js
         ? `${devBundleBase}${chunkNames.js}.bundle.dev.js`
         : `${devBundleBase}bundle.dev.js`;
       const jsChunk = _.find(assets.js, asset => _.includes(asset.chunkNames, chunkNames.js));
-      const cssChunk = _.find(assets.css, asset => _.includes(asset.chunkNames, chunkNames.css));
```
jchip: I am not sure what's the reason behind this line change here, and it seems it's a breaking change.

jshado1: Hmm, my earlier comment explaining why I did it seems to have been lost. I was asking what its purpose was because it seemed to actually have an adverse affect.

jshado1: I restored the filtering and confirmed it works both when `chunkNames.css` is an array or just a string.

jshado1:
@jchip Thanks for the PR feedback. Good call: `concat` is MUCH cleaner.

I added two cases for this update, but it's currently failing because branch coverage (99.56%) is < 100%; is there a branch coverage visualizer I can check to see what I missed?

jchip:
there is a file `coverage/lcov-report/index.html` that will show you all the missing coverage code.

jshado1:
@jchip based on the coverage report, [`if (data.criticalCSS);`](715/files#diff-ce6e041490b373a3a69267a9824a6c71R67) is the offender; however it **is** actually covered by a pre-existing case: `it("should inject critical css"`. I tried adding an additional case to assert that relevant`<link>` tags are _not_ added to the output, but it still claims the branch is not covered.

jchip:
Actually, look over the code, the `HEADER_BUNDLE_MARKER` should not consider critical CSS at all.  That's taken care of by the `CRITICAL_CSS_MARKER` hook.
So it can be removed.
Thanks.

jshado1:
@jchip updated and the test passes locally. Travis is having a stroke though. It looks like it encountered a network error. I don't have write access to the repo, so I don't get the rebuild button. SO suggests closing and re-opening the PR to trigger Travis again.

```
Network error fetching meta of clean-css: Cannot create property 'fynFo' on string
```

Under normal circumstances, why does the spec runner output that huge block of junk at the end? It makes it kinda difficult to figure out what happened.

I noticed this as the output was flying by:
>server
>    ✓ write some unit test for server code

Also, the hard-coded `'/js/'` thing I mentioned above is [`react-webapp.js:198`](https://github.com/electrode-io/electrode/blob/master/packages/electrode-react-webapp/lib/react-webapp.js#L198) (line 211 after my PR)
```js
const devBundleBase = `${devProtocol}${pluginOptions.devServer.host}:${pluginOptions.devServer
  .port}/js/`;
```

jshado1:
Travis is ✅ after the re-build from the close + re-open.

jchip:
thanks for the PR

jshado1:
Is there a release schedule I can check? Eager to start using this 😈

jchip:
could be today or within next few days.  @didi0613 has been doing some updates to the styles webpack config so once we verify that we will do a release.

jshado1:
@jchip I saw in #672 that React 16 support is coming soon. The release with this won't be tied to / require an upgrade to react 16, will it?

jchip:
electrode-react-webapp is not tied to react version.

PR #733, opened 2018-02-26
[patch] optionalRequire archetype/config folder for component archetype

didi0613:
Use `optionalRequire` for `archetype/config` folder, based on the `electrode-demo-index` instance. 

`electrode-demo-index` has no demo, its a template other components use for demo. the `archetype/config` is not a required folder and should have a default value for it.
Errors:
screen shot 2018-02-26 at 2 33 49 pm

### DIFF
--- a/packages/electrode-archetype-react-component-dev/config/webpack/partial/extract-style.js
+++ b/packages/electrode-archetype-react-component-dev/config/webpack/partial/extract-style.js
@@ -17,12 +17,7 @@ const sassLoader = require.resolve("sass-loader");
 const demoAppPath = Path.resolve(process.cwd(), "..", "..", "demo-app");
 const archetypeAppPath = Path.resolve(demoAppPath, "archetype", "config");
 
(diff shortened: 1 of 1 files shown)


packages/electrode-archetype-react-component-dev/config/webpack/partial/extract-style.js
```diff
@@ -14,12 +15,15 @@ const stylusLoader = require.resolve("stylus-relative-loader");
...
-  demoAppPath,
-  "archetype",
-  "config"
-);
-const archetypeAppWebpack = require(archetypeAppPath).webpack;
+const archetypeAppPath = Path.resolve(demoAppPath, "archetype", "config");
+
+const archetypeApp = optionalRequire(archetypeAppPath) || {
+  webpack: {
+    cssModulesSupport: undefined,
```
datvong-wm: `cssModuleSupport`  ..  but undefined is the same as {}

PR #747, opened 2018-03-15
[donotmerge][minor] enable i18n to archetype app

didi0613:
Tested in samples app: universal-react-app


packages/electrode-archetype-react-app/i18next/index.js
```diff
@@ -0,0 +1,28 @@
...
+  .use(reactI18nextModule)
+  .init({
+    detection: {},
+    fallbackLng: "en",
+    languages: ["en", "zh", "fr", "es", "it"],
+    react: {
+      wait: true
+    },
+    resources: {
+      en: { translation: require(`${__dirname}/../../../src/client/locales/en.json`) },
```
jchip: you can't use `__dirname` because that depends on where the module will be installed under `node_modules`

PR #761, opened 2018-03-23
fix error message display due to broken code check

jchip:


packages/electrode-webpack-reporter/src/client/components/warnings-errors.jsx
```diff
@@ -42,10 +42,11 @@ class WarningsErrors extends React.Component {
           </h2>
         </div>
 
-        {this.props.errors.map((e) => {
-          const errorHasExpandableCode = e.indexOf(")") > 0;
-          const errorTitle = errorHasExpandableCode ? e.slice(0, e.indexOf(")") + 1) : e;
-          if (errorHasExpandableCode) {
+        {this.props.errors.map((errStr) => {
+          const codeIdx = errStr.indexOf("<span");
+          const hasCode = codeIdx > 0;
```
datvong-wm: ">=" ?

jchip: do code only if there's a title.

PR #765, opened 2018-04-04
Add --inspect flag to clap server-watch

mcrowder65:
Might as well add this flag, this lets users open up chrome inspector to debug their server side code with breakpoints, etc.


mcrowder65:
For some reason the CLA thing is not going through :( 
image

PR #787, opened 2018-05-14
using `ReactDOM.hydrate` when content is rendered on the server

romanlv:
Using `render` or `hydrate` when possible

Updated dependencies of `electrode-redux-router-engine` to use React 16 when the application actually runs 

For PWA, since includes `react-notify-toast` (with React 15 direct dependency), react 16 must be listed in package.json to override it

### DIFF
--- a/packages/electrode-redux-router-engine/package.json
+++ b/packages/electrode-redux-router-engine/package.json
@@ -31,8 +31,8 @@
     "babel-preset-stage-0": "^6.0.15",
     "babel-register": "^6.5.2",
     "electrode-archetype-njs-module-dev": "^2.1.0",
-    "react": "^15.3.1 || ^0.14.8",
-    "react-dom": "^15.3.1 || ^0.14.8",
+    "react": "^16.0.0 || ^15.3.1 || ^0.14.8",
+    "react-dom": "^16.0.0 || ^15.3.1 || ^0.14.8",
     "xstdout": "^0.1.1"
   },
   "dependencies": {
@@ -43,10 +43,7 @@
     "react-router": "^3.0.0 || ^2.8.0",
     "redux": "^3.6.0"
   },
-  "peerDependencies": {
-    "react": "^16.0.0 || ^15.3.1 || ^0.14.8",
-    "react-dom": "^16.0.0 || ^15.3.1 || ^0.14.8"
-  },
+  "peerDependencies": {},
   "nyc": {
     "all": true,
     "reporter": [
--- a/packages/electrode-redux-router-engine/package.json
+++ b/packages/electrode-redux-router-engine/package.json
@@ -43,7 +43,10 @@
     "react-router": "^3.0.0 || ^2.8.0",
     "redux": "^3.6.0"
   },
-  "peerDependencies": {},
+  "peerDependencies": {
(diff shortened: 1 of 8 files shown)


packages/electrode-redux-router-engine/package.json
```diff
@@ -43,10 +43,7 @@
     "react-router": "^3.0.0 || ^2.8.0",
     "redux": "^3.6.0"
   },
-  "peerDependencies": {
-    "react": "^16.0.0 || ^15.3.1 || ^0.14.8",
-    "react-dom": "^16.0.0 || ^15.3.1 || ^0.14.8"
-  },
+  "peerDependencies": {},
```
jchip: any reason for removing this?

packages/generator-electrode/generators/app/templates/_package.json
```diff
@@ -42,6 +42,7 @@
     "koa-router": "^5.4.0",
     "koa-send": "^3.2.0",
     "koa-static": "^2.0.0", <% } if (isPWA) { %>
+    "react": "^16.3.2",
```
jchip: is the archetype not working w/o this?

samples/universal-react-node/package.json
```diff
@@ -32,6 +32,7 @@
   "dependencies": {
     "bluebird": "^3.4.6",
     "electrode-archetype-react-app": "../../packages/electrode-archetype-react-app",
+    "electrode-archetype-opt-react": "../../packages/electrode-archetype-opt-react",
```
jchip: is `electrode-archetype-react-app` not working w/o this?

romanlv: whenever `react-notify-toast` is present as a dependency it uses react `15.x`

packages/generator-electrode/generators/app/templates/_package.json
romanlv: same here, in my tests it didn't work, due to `react-notify-toast` dependency

samples/universal-react-node/package.json
romanlv: I guess it happens because `electrode-archetype-react-app` does not have `react` package as a direct dependency, but `react-notify-toast` does. Tested with `yarn` (and `npm`), not with `fyn`

jchip: Because we use `optionalDependencies` `yarn` doesn't work well.

yeah, the issue with react-notify-toast is known.  There's a PR https://github.com/jesusoterogomez/react-notify-toast/pull/55 for it.  It's merged, but not published yet.

jchip: I made the Electrode repo to run tests etc by pointing to local dependencies with `fyn` so the CI scripts automatically use `fyn`.

jchip: forgot to mention, by forcing react 16 at the top most level, it's just masking the issue with react-notify-toast, which pulls in react 15 regardless and you end up with both in your node_modules.  that's why I believe the proper fix is to update it in react-notify-toast.

romanlv:
@jchip, `react-notify-toas@0.4.1` is published, I removed direct `react 16` dependency

PR #788, opened 2018-05-15
export a main for package.json

jchip:
first of some updates.

New feature planned:

- Support dynamically switching the index.html template base on incoming request parameters etc

@romanlv @didi0613 @datvong-wm

### DIFF
--- a/packages/electrode-react-webapp/lib/hapi/index.js
+++ b/packages/electrode-react-webapp/lib/hapi/index.js
@@ -71,4 +71,9 @@ registerRoutesPlugin.attributes = {
   }
 };
 
-module.exports = registerRoutesPlugin;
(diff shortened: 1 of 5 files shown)


packages/electrode-react-webapp/lib/hapi/index.js
```diff
@@ -71,4 +71,9 @@ registerRoutesPlugin.attributes = {
   }
 };
 
-module.exports = registerRoutesPlugin;
+// registerRoutesPlugin.registerRoutes = registerRoutes;
+
+module.exports = {
```
datvong-wm: this definitely going to break peoples that use hapi directly.

jchip: It was always broken.

PR #789, opened 2018-05-21
unsupported methods returning 500s

nsunga:
unsupported method errors should return 405 codes.

currently returns a 500 error code.


jchip:
Hi thanks for the PR.  There're some minor linting errors:

```
lerna ERR! /home/travis/build/electrode-io/electrode/packages/electrode-redux-router-engine/lib/errors/unsupported-method-error.js
lerna ERR!   1:1   error  Use the global form of 'use strict'            strict
lerna ERR!   8:41  error  Newline required at end of file but not found  eol-last
```

PR #808, opened 2018-06-12
Use phantomjs as test browser for now till looper support headlesschrome

didi0613:
Since looper will support headless chrome soon, for now, just use phantomjs as karma browser inside generator.
Temporarily resolve the `Chrome_bin` cannot be found issue from looper.


didi0613:
no longer needed

PR #818, opened 2018-06-21
Run webpack dev middleware as part of app server

jchip:


romanlv:
@jchip what is the correct way to play with it? 
i'm trying `WEBPACK_DEV_MIDDLEWARE=true clap dev` from `samples/universal-react-node` and it cannot load js bundle, SS part seems to work initially, but does not reload after making changes

PR #824, opened 2018-06-26
[minor] Update component generator

didi0613:


jchip:
- please avoid formatting code with changes
- also, your format setting is not using 100 line width.  The exiting code is already formatted, just with 100 line width, and you reformatted it with 80 line width.

PR #839, opened 2018-06-28
Flow typed component example

didi0613:


jchip:
all these auto gen flow typed files look pointless.  Is there a way to not auto gen them?

didi0613:
@jchip did some researches, can u please clarify here:
`not auto gen them`, do u mean we can customize the libraries we need for flow instead of auto gen all of them?

Issue #852, opened 2018-07-08
Query strings not passed in during server side render (hapi)


devourment77:
## Type of issue
Bug

## My environment
OS version/details: MacOS 10.13.5 / Chrome Version 67.0.3396.87 (Official Build) (64-bit)
Node version: 10.3.0 (run node --version in your terminal)
npm version: 6.1.0 (run npm --version in your terminal)

## Expected behavior
Query strings to be passed into to `props.location.search` and / or `props.history.location.search` during server side rendering.  This is with the new react router v4 version.

## Current behavior
They are **not** being passed in, only the path (at least for hapi).

It looks to be an issue on this line in `redux-router-engine.js`:
```
const location = req.path || (req.url && req.url.path);
```

Ideally (at least for hapi), req.url.path will be first as it will pass in the query string (tested this and it works).

## Steps to reproduce the behavior
- create a page using a query string.  EX: `/some-page?browse=a`
- on server side render, location.search is not set.

PR #854, opened 2018-07-08
[minor] pass location with query string to routes

jchip:
fix https://github.com/electrode-io/electrode/issues/852

- the `location` passed to router should be an object similar to what `url.parse` returns, that is similar to client side behavior.


didi0613:
lgtm

PR #867, opened 2018-07-18
[patch] Fix above the fold path inside universal-react-node sample

didi0613:
skip true
skip true

skip false
skip false


jchip:
please verify with `skip=true`?

jchip:
Please show page source to verify that content below the fold are:

- Not rendered when skip=true
- Rendered when skip=false

didi0613:
@jchip 
screen shot 2018-07-19 at 12 10 35 pm

jchip:
`skip` is supposed to be passed in as `props`

jchip:
The `skip` flag is only needed to be processed on the server side.

- remove `query-string` and do not do anything with it in the component.
- server side process the `skip` flag, set it in Redux initial state.
- pass the `skip` flag into the components through redux initial state

PR #885, opened 2018-07-30
[minor][feat] add basename support for SSR Links

Bangsheng:
If you use Link from react-router-dom with a basename, for example,
basename: '/my-base' and a `<Link to="/to-target">Test</Link>`

CSR will render it as `<a href="/my-base/to-target">Test</a>`
But SSR will render it as `<a href="/to-target">Test</a>` because the render engine doesn't support basename. 

This PR is to add basename support to the render engine, so we can get same render result CSR vs. SSR.


Bangsheng:
@jchip Thanks for merging this! May I know when will you publish this feature?

PR #929, opened 2018-09-12
[patch][bug] handle text based scripts

PeoB:
As they end up with the tag `<scriptundefined>` at the moment which seems unintended


jchip:
the bug is actually here https://github.com/electrode-io/electrode/blob/master/packages/electrode-react-webapp/lib/react/token-handlers.js#L159

it should've passed in `context.user.scriptNonce`

PR #931, opened 2018-09-12
[minor][feat] Update _matchRoute to support match with basename

Bangsheng:
The issue I want to fix is described as below:
my routes:
```
const routes = [
  { component: Root,
    routes: [
      { path: '/',
        exact: true,
        component: Home
      },
      { path: '/child/:id',
        component: Child,
        routes: [
          { path: '/child/:id/grand-child',
            component: GrandChild
          }
        ]
      }
    ]
  }
]
```

And I want to serve my app with a basename `/my-sub-app`. When user go to https://myhost.com/my-sub-app/child/2 the current `_matchRoute` function will match against `req.pathname` which is `/my-sub-app/child/2`. That will result in a mis match.


jchip:
thanks for PR.  merged manually.

Bangsheng:
@jchip Thanks for the awesome work and merging this PR!

PR #989, opened 2018-10-10
Hapi17

datvong-wm:


datvong-wm:
major.  The OSS doesn't have dual support.  It only use Hapi 17 plugin.

PR #998, opened 2018-10-17
[minor][bug] fixes CEECORE-449 test files are copied to lib from src

snekse:
from https://stackoverflow.com/questions/52629761/electrode-keeping-jest-test-files-out-of-lib-on-build

Since we babel src to lib directly, and if user put Jest tests in src also, then we are copying them to lib.

So looks like we need some filtering to nudge babel to skip test files.


jchip:
Thanks for PR and creating the sample app. 👍

Issue #1077, opened 2018-11-28
[WEBPACK 4] using `import "~..."` in css files errors out


Anujan:
## Type of issue
Bug
----

## My environment

* *OS version/details*: `MacOS `
* *Node version:* `8.12.0` (run `node --version` in your terminal)
* *npm version:* `6.4.1` (run `npm --version` in your terminal)

## Expected behavior

Electrode React Web App v6 builds externally referenced CSS files 

## Current behavior

When we use `@import "~normalize.css/normalize.css";`, webpack errors out saying it's unable to find `"~normalize.css/normalize.css"`. This was working in Electrode React Web App v5

## Command line output

```
ERROR in ./client/styles/globals.css
Module build failed (from ../node_modules/css-loader/index.js):
ModuleBuildError: Module build failed (from ../node_modules/postcss-loader/src/index.js):
Error: Failed to find '~normalize.css/normalize.css'
    in [
        /Users/a0p00tg/Repos/productpage-webapp/packages/product-electrode/src/client/styles
    ]
    at /Users/a0p00tg/Repos/productpage-webapp/node_modules/postcss-import/lib/resolve-id.js:61:11
    at <anonymous>
    at runLoaders (/Users/a0p00tg/Repos/productpage-webapp/packages/product-electrode/node_modules/webpack/lib/NormalModule.js:286:20)
    at /Users/a0p00tg/Repos/productpage-webapp/node_modules/loader-runner/lib/LoaderRunner.js:364:11
    at /Users/a0p00tg/Repos/productpage-webapp/node_modules/loader-runner/lib/LoaderRunner.js:230:18
    at context.callback (/Users/a0p00tg/Repos/productpage-webapp/node_modules/loader-runner/lib/LoaderRunner.js:111:13)
    at Promise.resolve.then.then.catch (/Users/a0p00tg/Repos/productpage-webapp/packages/product-electrode/node_modules/postcss-loader/src/index.js:208:9)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
 @ ./client/styles/globals.css

ERROR in ./client/styles/globals.css (../node_modules/css-loader??ref--5-2!../node_modules/postcss-loader/src??postcss!./client/styles/globals.css)
Module build failed (from ../node_modules/postcss-loader/src/index.js):
Error: Failed to find '~normalize.css/normalize.css'
    in [
        /Users/a0p00tg/Repos/productpage-webapp/packages/product-electrode/src/client/styles
[...]

PR #1078, opened 2018-11-29
demo CSS @import from node_modules

jchip:
demo solution for https://github.com/electrode-io/electrode/issues/1077


didi0613:
lgtm

PR #1081, opened 2018-11-30
[patch] webpack 4: concatenateModules false

didi0613:
concatenateModules flag is to find segments of the module graph which can be safely concatenated into a single module.

This is not enabled under dev mode with our previous webpack 3 config. It is newly added for improving performance with webpack v4.  However, it causes some issues when testing with home app from `@walmart/feds-wmt-header` module. 

Change back to default: `concatenateModules: false` in dev and `concatenateModules: true` in prod.


jchip:
how is it enabled for prod?

didi0613:
@jchip Default behavior: Enabled in production mode. Elsewise disabled.

PR #1111, opened 2019-01-09
[minor][bug] fix server watch when using typescript

Anujan:
* TypeScript extensions added to nodemon command
* Removed the webpack-dev-middleware condition that made nodemon not watch the server folder. The server folder isn't built with webpack, so it wasn't being watched by anything.

### DIFF
--- a/packages/electrode-archetype-react-app/arch-clap.js
+++ b/packages/electrode-archetype-react-app/arch-clap.js
@@ -684,7 +684,11 @@ Individual .babelrc files were generated for you in src/client and src/server
       dep: [".init-bundle.valid.log"],
       desc: "Start app's node server in watch mode with nodemon",
       task: function() {
-        const watches = [Path.join(eTmpDir, "bundle.valid.log"), AppMode.src.server, "config"]
+        const watches = (archetype.webpack.devMiddleware
+          ? []
+          : [Path.join(eTmpDir, "bundle.valid.log")]
+        )
+          .concat(["config", AppMode.src.server])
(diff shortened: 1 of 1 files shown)


packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -684,11 +684,7 @@ Individual .babelrc files were generated for you in src/client and src/server
       dep: [".init-bundle.valid.log"],
       desc: "Start app's node server in watch mode with nodemon",
       task: function() {
-        const watches = (archetype.webpack.devMiddleware
-          ? []
-          : [Path.join(eTmpDir, "bundle.valid.log"), AppMode.src.server]
-        )
-          .concat("config")
+        const watches = [Path.join(eTmpDir, "bundle.valid.log"), AppMode.src.server, "config"]
```
jchip: The reason for the `devMiddleware` condition is due to HMR for SSR.

When HMR is enabled with the `dev-middleware`, it is detecting files changes through webpack and dynamically reload them in the same node process.

If you make `nodemon` watch them while in `devMiddleware` mode, nodemon will **restart** the node process any time webpack recompiles and updates `bundle.valid.log` file.

However, it may make sense that `AppMode.src.server` should always be watched though since code there isn't subjected to webpack and nodemon should restart the server.  The drawback of doing this would be HMR is disrupted and you have to reload the browser if nodemon restarts the node process, but if you changed server code, maybe that's what should happen anyways.

Issue #985, opened 2018-10-09
Jest custom root file path not working.


sagar-prolix:
# Jest custom root file path not working.

### config
```sh
module.exports = {
  options: {
    karma: false
  },
  jest: {
    verbose: true,
    testEnvironment: "node",
    setupTestFrameworkScriptFile: "<rootDir>/jest-setup.js",
    globals: {},
    moduleNameMapper: {
      "\\.(css|scss|less)$": "<rootDir>/__mock__/styleMock.js",
      "\\.(png|jpg|gif|ttf|eot|svg)$": "<rootDir>/__mock__/fileMock.js"
    },
    roots: [
        "<rootDir>/jest-test"
    ]
  }
};
```

# issue
electrode-archetype-react-app/arch-clap.js
```sh
".jest.test-frontend-cov": () => {
      const srcJestFiles = glob.sync(`${Path.resolve(AppMode.src.dir)}/**/\*.{test,spec}.{js,jsx}`);

      if (shell.test("-d", "_test_") || srcJestFiles.length > 0) {
        logger.info("Running jest unit tests");
        return mkCmd(`~$jest`, `--config ${archetype.config.jest}/jest.config.js`);
      }
      return undefined;
    }
```
# Description
- srcJestFiles variable always check "src" folder for jest test file not check custom root path.
 - if you give custom file for jest and add blank file in scr folder anyware it will trigger jest command otherwise it will return undefined.
 - It should check jest root path.

PR #1145, opened 2019-02-04
[patch][chore] read custom jest config

1846689910:


packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -754,24 +754,27 @@ Individual .babelrc files were generated for you in src/client and src/server
     "test-frontend-ci": ["?.karma.test-frontend-ci"],
 
     ".jest.test-frontend-cov": () => {
+      let runJest;
       const testDir = ["_test_", "__tests__"].find(x => shell.test("-d", x));
-
-      const srcJestFiles =
-        !testDir &&
-        glob.sync(`**/*.{test,spec}.{js,jsx,ts,tsx}`, {
+      if (! testDir) {
```
jchip: There are three scenarios that we want to run jest test.
First: If user has a directory that's named `_test_` or `__tests__`.   
Second: If any files name matches `*.{test,spec}.{js,jsx,ts,tsx}`
Third: Users has their own config to identify tests, in which case we just run jest.

jchip: Note that if **any** of the scenarios is true independently, we run jest test.

packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -755,23 +755,24 @@ Individual .babelrc files were generated for you in src/client and src/server
...
-
-      if (testDir || srcJestFiles.length > 0) {
-        if (testDir) {
-          makeBabelRc(Path.join(testDir, "client"), "babelrc-client.js");
-        }
+        }).length > 0;
+      }
+      if (runJest) {
+        const { roots } = archetype.jest;
+        runJest = roots && roots.map(x => x.replace("<rootDir>", process.cwd()));
```
jchip: what's the `replace` for?

jchip: more specifically, what's using map/replace to process the array for?  The value of that array is not used anywhere.

jchip:
fix https://github.com/electrode-io/electrode/issues/985

PR #1152, opened 2019-02-15
change command to lower case

lifengli:
to match its actual behavior


jchip:
thank you. I prefer looking at the command keys in upper case. :) 
Isn't it better to make the input handler to convert `str` to lowercase so it's case insensitive?  
Note that `str` could be `undefined`.

jchip:
@lifengli Hi, I want to check if you will be able to update this soon?

lifengli:
> @lifengli Hi, I want to check if you will be able to update this soon?

Hi, please give me sometime since I'm new to this codebase. If you've already fixed it, feel free to publish it. Thanks,

lifengli:
updated the PR, please review

PR #1166, opened 2019-03-07
server side bundle selection for different targets

1846689910:


packages/electrode-archetype-react-app-dev/config/babel/babelrc-client.js
```diff
@@ -33,25 +33,6 @@ const basePlugins = [
     "@babel/plugin-proposal-class-properties",
     { loose: looseClassProps }
   ],
-  "@babel/plugin-transform-object-super",
```
jchip: remove this in another PR

packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -498,9 +497,40 @@ function makeTasks(xclap) {
...
       task: () => {
-        if (buildDistDirs.length > 0) {
-          return exec(`mv`, ...buildDistDirs, `dist`);
-        }
+        buildDistDirs.forEach(dir => {
+          scanDir
+            .sync({
+              dir,
+              includeRoot: true,
+              filter: file => /.+(\.sw-registration|-main\.bundle)\.js(\.map)?$/.test(file)
```
jchip: add comment for the regex

packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -619,7 +649,7 @@ Individual .babelrc files were generated for you in src/client and src/server
     "check-dev": ["lint", "test-dev"],
 
     clean: [".clean.dist", ".clean.lib", ".clean.prod", ".clean.etmp", ".clean.dll"],
-    ".clean.dist": () => shell.rm("-rf", "dist"),
+    ".clean.dist": () => shell.rm("-rf", ...shell.ls("-d", "dist*")),
```
jchip: why not use `buildDistDirs`?

1846689910: remove all the `dist-X` bundles, even `buildDistDirs` are different for the next time

jchip: you must only worry about the target being built.  removing all you will have multiple builds stepping on each other.

packages/electrode-react-webapp/lib/react-webapp.js
```diff
@@ -4,6 +4,16 @@ const _ = require("lodash");
 const Path = require("path");
 const assert = require("assert");
 const AsyncTemplate = require("./async-template");
+const filterScanDir = require("filter-scan-dir");
+const otherStats = filterScanDir.sync({
+  dir: Path.resolve("dist", ".."),
+  includeRoot: true,
+  filter: file => file === "stats.json"
```
jchip: you are only looking for a single file `stats.json`?

jchip: if you are just looking for a single file, why do you need scan dir?

packages/electrode-react-webapp/lib/react/token-handlers.js
```diff
@@ -71,14 +72,25 @@ module.exports = function setup(handlerContext /*, asyncTemplate*/) {
       : `${prodBundleBase}${assets.manifest}`;
   };
 
+  const getBundleJsNameByQuery = data => {
+    let { name } = data.jsChunk;
+    const { __dist } = data.query;
```
divyamnpll: so do we need to set the `_dist` in query parameter to match the babel target  ?

1846689910: This is a temporary way to explicitly use specific bundle for debug the feature. Finally, it will depends on how you want to detect the bundle you want to use

PR #1169, opened 2019-03-09
Remove PWA optional check

swashcap:
### Problem

The [PWA optional check](https://github.com/electrode-io/electrode/blob/3c54ce0e0742b39f101158d0c5cc2474cc5978ff/packages/electrode-archetype-opt-pwa/optional-check.js#L38) exits with `>0` when the archetype config has `"pwa": false`.

### Solution

npm appears to rollback installs when a package's `preinstall` script fails. Remove the `process.exit(1)` line so the install works.

### Testing

No tests included! Not sure about testing it.

### DIFF
--- a/packages/electrode-archetype-opt-pwa/optional-check.js
+++ b/packages/electrode-archetype-opt-pwa/optional-check.js
@@ -35,7 +35,6 @@ try {
     console.log(
       `${name}: skipping install because archetype/config set options.${configName} to false`
     );
-    process.exit(1);
   }
 } catch (e) {}
 


jchip:
`electrode-archetype-react-app-dev` pulls `electrode-archetype-opt-pwa` as `optionalDependencies`.  By exiting with status 1 is how the module tells npm to "ignore it" through `optionalDependencies`.  

When setting `pwa` to `true` and it exits 0, npm then continue to install it and its dependencies.

PR #1188, opened 2019-03-20
back compatible with no target specified

1846689910:
+ bring back hash in filename
+ default bundle filename and chunkfilename performs the same as before. When no `BABEL_ENV_TARGETS` specified, compatible with before
+ css file still without hash


packages/generator-electrode/generators/app/templates/src/client/components/demo-dynamic-import.jsx
```diff
@@ -13,7 +13,7 @@ const Fallback = () => (
   </div>
 );
 
-let Demo = loadable(() => import(/* webpackChunkName: "fake" */ "./demo-fake"), {
+let Demo = loadable(() => import("./demo-fake"), {
```
jchip: what's this change for?

1846689910: remove the chunk for this component,  and make fewer generated js files

1846689910: component name changed and webpackChunkName brought back

PR #1193, opened 2019-03-29
fix entry format

1846689910:
with user defined `entry.config.js` exists and `enableBabelPolyfill` as `true`, the webpack will throw error for invalid entry format


1846689910:
close and merge with https://github.com/electrode-io/electrode/pull/1194

PR #1196, opened 2019-04-01
[MINOR]make the babel config configurable 

divyamnpll:
This allows apps to override the babel exclude and babelrc option which enables apps to transpile dependencies in node_modules

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/archetype.js
+++ b/packages/electrode-archetype-react-app-dev/config/archetype.js
@@ -51,6 +51,10 @@ const webpackConfigSpec = {
     env: "ELECTRODE_LOAD_DLLS",
     type: "json",
     default: {}
+  },
+  extendBabelLoader: {
+    type: "json",
+    default: {}
   }
 };
 
@@ -80,6 +84,9 @@ const babelConfigSpec = {
     env: "ENV_TARGET",
     type: "string",
     default: "default"
+  },
+  noBabelRc: {
+    default: false
   }
 };
 
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js
@@ -18,7 +18,7 @@ module.exports = function(options) {
 
   const test = archetype.babel.enableTypeScript ? /\.[tj]sx?$/ : /\.jsx?$/;
 
-  let customBabelConfig = {};
+  const customBabelConfig = {};
   if (archetype.babel.noBabelRc) {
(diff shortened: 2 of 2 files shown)


jchip:
this is not necessary because babel config is configured by your own `.babelrc` or `babelrc.js` file, just don't extend the archetype's babelrc in yours.

divyamnpll:
@jchip we want the ability to override the babelexclude option https://github.com/electrode-io/electrode/blob/42c9e4cc8b6cd6eeec192800c894f3293ed768b2/packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js#L12 
 and want to set babelrc option as false https://babeljs.io/docs/en/options#babelrc as we want to transpile the node_modules also based on target browser

jchip:
none of your changes would allow you to override babel-loader's options, including `exclude`?

divyamnpll:
there is no other way to configure babel exclude unless if there is a way if `options` can be configurable from app hhttps://github.com/electrode-io/electrode/blob/master/packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js#L10

jchip:
there are multiple changes in this PR that doesn't align with the stated goal.
- If you want to override babel configs, that can already be achieved through your own `babelrc` files.
-- therefore, I am not sure why you need the `noBabelRc` addition because you just don't add a `babelrc`

And if your goal is to be able to configure exclude, none of your changes would enable you to do that.

divyamnpll:
@jchip babelexclude can be configured by setting the options in `archetype.webpack.extendBabelLoader` https://github.com/electrode-io/electrode/blob/42c9e4cc8b6cd6eeec192800c894f3293ed768b2/packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js#L37

Currently this is not possible because extendBabelLoader is not defined in the default config
https://github.com/electrode-io/electrode/pull/1196/files#diff-8dbbb912183342671705d3c7d77702ebR55

divyamnpll:
`If you want to override babel configs, that can already be achieved through your own babelrc files.`
Agree we can override the babel config using own babelrc files , but 
our goal is to have not just override babel config but to disable babelrc look up behavior https://babeljs.io/docs/en/options#babelrc, Please let us know if its possible to disable the lookup behavior in archetype with any other possible configuration

divyamnpll:
I am closing this  PR as we need these changes in V5

and this PR  https://github.com/electrode-io/electrode/pull/1197 brings the intended noBabelrc change
https://github.com/electrode-io/electrode/pull/1197/files#diff-32b4a8f309f74f9c3cf25155f5e4f8c2R27

PR #1208, opened 2019-04-10
add extendLoader to archetype babel config

1846689910:


packages/electrode-archetype-react-app-dev/config/archetype.js
```diff
@@ -68,6 +64,10 @@ const babelConfigSpec = {
     env: "ENV_TARGET",
     type: "string",
     default: "default"
+  },
+  extendLoader: {
```
jchip: add a comment to indicate that this is only used if multi targets is used.

PR #1226, opened 2019-04-23
use core-js and regenerator-runtime to replace @babel/polyfill

1846689910:


1846689910:
open with https://github.com/electrode-io/electrode/pull/1227

PR #1240, opened 2019-05-13
dynamically imported components server side rendering

1846689910:
change the dynamic imported components from CSR to SSR
current source elements:
Screen Shot 2019-05-13 at 1 21 24 PM
after change, the source elements will contain content of loadable components:
Screen Shot 2019-05-13 at 1 21 58 PM


packages/electrode-redux-router-engine/lib/redux-router-engine.js
```diff
@@ -263,21 +263,23 @@ class ReduxRouterEngine {
       } else {
         ssrApi = withIds ? ReactDomServer.renderToString : ReactDomServer.renderToStaticMarkup;
       }
-
+      const extractor = util.getExtractor(Path.resolve("./dist/js/loadable-stats.json"));
```
jchip: this is a hot code path, and calling `util.getExtractor`, which invokes fs sync APIs is not a good idea.

- Is a `new ChunkExtractor` needed every time?  Can it be done only once for production.
- note for that non-production a new one is needed every time in case of dev updates.

1846689910: solved, production will keep using single extractor, but dev will create new one every time

packages/electrode-archetype-react-app-dev/lib/dev-admin/middleware.js
```diff
@@ -214,6 +218,15 @@ class Middleware {
 
     const transferIsomorphicAssets = (fileSystem, cb) => {
       const isoConfig = loadIsomorphicConfig();
+      const loadableStats = `/${LOADABLE_STATS}`;
+
+      if (fileSystem.existsSync(loadableStats)) {
+        const source = fileSystem.readFileSync(loadableStats);
+        const dir = Path.resolve("./dist/js");
+        if (!Fs.existsSync(dir)) shell.mkdir("-p", dir);
+        Fs.writeFile(`${dir}${loadableStats}`, source, cb);
```
jchip: the function `transferIsomorphicAssets` now could end up invoking the callback `cb` twice.

packages/electrode-redux-router-engine/lib/util.js
```diff
@@ -10,5 +14,14 @@ module.exports = {
     if (p.startsWith("/")) return p;
 
     return p.startsWith(".") ? Path.resolve(baseDir || "", p) : p;
+  },
+  getExtractor: () => {
+    if (process.env.NODE_ENV === "production" && extractor) return extractor;
```
jchip: It may be cleaner if the function name `getExtractor` is called `createdStatExtractor` and it doesn't deal with memoization for `production`.  Instead, memoization should be handled by caller.

packages/electrode-redux-router-engine/lib/util.js
```diff
@@ -1,6 +1,9 @@
 "use strict";
 
 const Path = require("path");
+const Fs = require("fs");
+const { ChunkExtractor } = require("@loadable/server");
+const statsFile = Path.resolve("./dist/js/loadable-stats.json");
```
jchip: save this to `dist/server`

packages/electrode-redux-router-engine/lib/util.js
```diff
@@ -10,5 +12,22 @@ module.exports = {
...
+    return Fs.existsSync(statsFile)
+      ? new ChunkExtractor({ statsFile })
+      : {
+          collectChunks: app => app
+        };
+  },
+  getTargetByQuery: query => {
+    const __dist = query && query.__dist;
+    let target = "default";
+    if (__dist && Fs.existsSync(Path.resolve(`dist-${__dist}`))) {
```
jchip: not a good idea to use `fs.existsSync` in hot code path

PR #1241, opened 2019-05-13
migrate uglify to terser

AndyWMT:
this replaces uglify with terser as uglify-es is no longer maintained and uglify-js does not support ES6+.

let me know if I'm missing anything, anything being cleaned out needs putting back ?

is hasMultiTargets, INSPECTPACK_DEBUG , archetype.webpack.minify being used or can we nuke it ?

- if prod use terser to uglify 
- if dev don't run any sort of uglification on code (this speeds stuff up considerably)
- I added the flag to preserve comments for license etc ..

extension of https://github.com/electrode-io/electrode/pull/1238

### DIFF
--- a/packages/electrode-archetype-react-app-dev/package.json
+++ b/packages/electrode-archetype-react-app-dev/package.json
@@ -115,6 +115,7 @@
     "sugarss": "^1.0.1",
     "sw-precache": "^5.0.0",
     "sw-toolbox": "^3.4.0",
+    "terser-webpack-plugin-legacy": "^1.2.3",
     "uglify-js": "^3.0.26",
     "uglifyjs-webpack-plugin": "^1.2.2",
     "url-loader": "^0.6.2",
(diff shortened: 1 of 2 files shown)


packages/electrode-archetype-react-app-dev/package.json
```diff
@@ -115,6 +115,7 @@
     "sugarss": "^1.0.1",
     "sw-precache": "^5.0.0",
     "sw-toolbox": "^3.4.0",
+    "terser-webpack-plugin-legacy": "^1.2.3",
     "uglify-js": "^3.0.26",
```
DenysMikhalenkoWalmart: Do we still need these?

AndyWMT: looks like its used here https://github.com/electrode-io/electrode/blob/v5.x.x/packages/electrode-archetype-react-app-dev/config/webpack/plugins/sw-precache.js#L22 
but uglifyjs-webpack-plugin can go I think

1846689910: Hi AndyWMT, why `use strict` and `minify` flag are removed

AndyWMT: @1846689910  correct me but as I understand it node js wraps modules by default so 'use strict' is redundant in this case. 

there is no need for the minify as it is not using uglify in dev

AndyWMT:
closing but please implement on next archetype

PR #1244, opened 2019-05-14
[minor][feat] support template written in JSX

jchip:
TBD: docs

Instead of using a simple string tokenizer template, utilize React's JSX standard for creating the index page template, gaining all the advantages of writing in JSX.

Not React.

Legacy tokens for the template fully supported.

### DIFF
--- a/packages/electrode-react-webapp/lib/jsx/JsxRenderer.js
+++ b/packages/electrode-react-webapp/lib/jsx/JsxRenderer.js
@@ -43,7 +43,7 @@ class JsxRenderer {
     return this._options.insertTokenIds;
   }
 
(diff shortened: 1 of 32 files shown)


packages/electrode-react-webapp/lib/jsx/JsxRenderer.js
```diff
@@ -0,0 +1,249 @@
...
+      options
+    );
+    this._template = options.template;
+  }
+
+  get insertTokenIds() {
+    return this._options.insertTokenIds;
+  }
+
+  get tempalteFullPath() {
```
DenysMikhalenkoWalmart: typo

PR #1249, opened 2019-05-18
set babel env target for dev mode with archetype v5

1846689910:
define `envTargets` in `archetype/config/index.js`
use `clap dev --target=es6` or alias `clap dev-es6`

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js
@@ -11,7 +11,6 @@ module.exports = function(options) {
   const { options: babelLoaderOptions = {}, ...rest } = archetype.babel.extendLoader;
   const isProduction =
     process.env.BABEL_ENV === "production" || process.env.NODE_ENV === "production";
-  const isDevelopment = process.env.NODE_ENV === "development";
   const getBabelrcClient = () => {
     const babelrcClient = JSON.parse(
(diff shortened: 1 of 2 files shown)


packages/electrode-archetype-react-app-dev/config/webpack/partial/babel.js
```diff
@@ -25,7 +26,7 @@ module.exports = function(options) {
         targets,
         useBuiltIns: "entry",
         corejs: "2",
-        modules: isProduction ? false : "commonjs"
+        modules: isProduction || (isDevelopment && target !== "default") ? false : "commonjs"
```
jchip: I don't get what's going on here.  Could you explain it for me please?

packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -669,6 +685,12 @@ Individual .babelrc files were generated for you in src/client and src/server
       }
     },
 
+    "dev-es6": {
+      desc:
+        "Start your app with watch in development mode with webpack-dev-server with ENV_TARGET=es6",
+      task: ["dev-target-es6", "dev"]
```
jchip: `task: "dev --target=es6"` should work also right?  No need for the extra `dev-target-es6` task.

jchip: correction: `task: ["dev --target=es6"]`

Issue #1255, opened 2019-05-24
MiniCssExtractPlugin css/sass image prefix bug


brandonburkett:
## Type of issue

Bug

----

## My environment

* *OS version/details*: `macOS 10.14.4`
* *Node version:* `12.3.1` 
* *npm version:* `6.9.0`

## Expected behavior
Images included via css and sass previously worked with ExtractTextPlugin, but with the switch to MiniCssExtractPlugin, the seem to be prefixed incorrectly and missing `/js/`

## Current behavior
CSS / SASS image includes to be prefixed by`/js/`

## Steps to reproduce the behavior
* spin up electrode app via ignite
* add the following to custom.css with a test image
```
.test {
  background-image: url("./test.jpg");
  background-color: #cccccc;
}
```

* add the folowing to home.jsx `<div styleName={"custom.container custom.test"}>`
* get a 500 error for image
** actual (broken): http://127.0.0.1:2992/fd2b67abf0e177d206a4e6f45a5b8eca.jpg
** what it should be (works): http://127.0.0.1:2992/js/fd2b67abf0e177d206a4e6f45a5b8eca.jpg

1846689910:
Thanks for posting issue, this was resolved by https://github.com/electrode-io/electrode/pull/1261

brandonburkett:
No prob at all, `MiniCssExtractPlugin` supports hot reloading now!  Would love it was added, then I could drop my custom extract styles config.  EX:

```
if (archetype.options.sass) {
    rules.push({
      _name: `extract${cssModuleSupport ? '-css' : ''}-scss`,
      test: /\.(scss|sass)$/,

      /* CUSTOM - live reload on dev */
      use: [
        {
          loader: MiniCssExtractPlugin.loader, options: {
            publicPath: '/js/',
            // only enable hot in development
            hmr: process.env.NODE_ENV === 'development',
            // if hmr does not work, this is a forceful method.
            reloadAll: true,
          },
        },
        ...(cssModuleSupport
          ? [cssModuleQuery, postcssQuery, sassQuery]
          : [cssQuery, postcssQuery, sassQuery]),
      ],
    });
  }
```

PR #1262, opened 2019-05-29
enable styles HMR

1846689910:
fix the subsequent request under https://github.com/electrode-io/electrode/issues/1255

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract-style.js
+++ b/packages/electrode-archetype-react-app-dev/config/webpack/partial/extract-style.js
@@ -14,6 +14,7 @@ const cssLoader = require.resolve("css-loader");
 const stylusLoader = require.resolve("stylus-relative-loader");
 const postcssLoader = require.resolve("postcss-loader");
 const lessLoader = require.resolve("less-loader");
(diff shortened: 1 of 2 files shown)


packages/electrode-archetype-react-app-dev/lib/dev-admin/middleware.js
```diff
@@ -229,9 +229,11 @@ class Middleware {
       }
```
jchip: changes in this file in a separate PR please

packages/electrode-archetype-react-app-dev/config/webpack/partial/extract-style.js
```diff
@@ -14,6 +14,7 @@ const cssLoader = require.resolve("css-loader");
 const stylusLoader = require.resolve("stylus-relative-loader");
 const postcssLoader = require.resolve("postcss-loader");
 const lessLoader = require.resolve("less-loader");
+const isDevelopment = process.env.NODE_ENV === "development";
```
jchip: `!== "production"`

jchip: is this flag set in archetype's config?

PR #1264, opened 2019-05-30
fix `isomorphic-assets.json` write path

1846689910:


packages/electrode-archetype-react-app-dev/lib/dev-admin/middleware.js
```diff
@@ -229,9 +229,11 @@ class Middleware {
       }
 
       if (isoConfig.assetsFile) {
-        const assetsFile = Path.resolve(isoConfig.assetsFile);
+        const assetsFile = Path.join(this.memFsCwd, isoConfig.assetsFile);
         const source = fileSystem.readFileSync(assetsFile);
-        Fs.writeFileSync(assetsFile, source);
+        const dir = Path.resolve("./dist");
+        if (!Fs.existsSync(dir)) shell.mkdir("-p", dir);
+        Fs.writeFileSync(Path.join(dir, assetsFile), source);
```
jchip: not a good idea to use `assetsFile` which was joined with memFs's CWD.  Real FS is always `process.cwd()/dist/server`.

PR #1295, opened 2019-07-12
[patch] [bug] CEECORE-686, Fixed broken Windows memfs page

scrajesh:
* Pass Posix Path methods/props when serve-index-fs uses MemoryFileSystem
* Also the webpack middleware should use Posix Path methods/props when using MemoryFileSystem


packages/electrode-archetype-react-app-dev/lib/dev-admin/middleware.js
```diff
@@ -149,12 +150,12 @@ class Middleware {
...
     this.listAssetPath = urlJoin(this.publicPath, "/");
 
-    // webpack dev middleware's memfs root is where the actual output from webpack
     this.memFsCwd = this.devMiddleware.fileSystem.existsSync(process.cwd()) ? process.cwd() : "/";
     this.cwdMemIndex = serveIndex(this.memFsCwd, {
       icons: true,
       hidden: true,
-      fs: this.devMiddleware.fileSystem
+      fs: this.devMiddleware.fileSystem,
+      path: isWin32 ? Path.posix : null
```
jchip: there is no need for the win32 check.  `Path.posix` is always available on all platforms, and we need to use that for memfs on all platforms.

packages/electrode-archetype-react-app-dev/lib/dev-admin/middleware.js
```diff
@@ -294,10 +295,11 @@ doReload(1); </script></body></html>`)
       );
     }
 
-    const serveStatic = (baseUrl, fileSystem, indexServer, cwd) => {
+    const serveStatic = (baseUrl, fileSystem, indexServer, cwd, isMemFs) => {
       req.originalUrl = req.url; // this is what express saves to, else serve-index nukes
-      req.url = req.url.substr(baseUrl.length) || "/";
-      const fullPath = Path.join(cwd || process.cwd(), req.url);
+      req.url = req.url.substr(baseUrl.length) || cwd;
+      const PathLib = isWin32 && isMemFs ? Path.posix : Path;
```
jchip: again, no win32 check needed.  `isMemFs` should always use `Path.posix`.

packages/electrode-archetype-react-app-dev/lib/dev-admin/middleware.js
scrajesh: removed Win32 condition, using posix() directly whenever MemFS is involved irrespective of the platform.

packages/electrode-archetype-react-app-dev/lib/dev-admin/middleware.js
scrajesh: removed Win32 condition, using posix() directly whenever MemFS is involved irrespective of the platform.

PR #1297, opened 2019-07-17
CEECORE-773: App Archetype import scss in SSR

divyakarippath:
Added "scss" and "styl" extensions and preprocessors  to cssModuleHook to support scss/styl imports  with SSR


jchip:
This only works if CSS Module is enabled.
The other case is when CSS Module is disabled but user still import `*.styl`, `*.scss` in order to get mini-css-extract-plugin to process them.

PR #1305, opened 2019-07-26
[feat] add `enableDynamicImport` to archetype

1846689910:
+ set a flag for dynamic import babel setup and webpack plugins
+ fix missing chunks for dynamic import syntax

### DIFF
--- a/packages/electrode-archetype-react-app-dev/config/archetype.js
+++ b/packages/electrode-archetype-react-app-dev/config/archetype.js
@@ -74,7 +74,7 @@ const karmaConfigSpec = {
 
 const babelConfigSpec = {
   enableTypeScript: { env: "ENABLE_BABEL_TYPESCRIPT", default: false },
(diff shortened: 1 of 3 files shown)


packages/electrode-archetype-react-app-dev/config/archetype.js
```diff
@@ -74,6 +74,7 @@ const karmaConfigSpec = {
 
 const babelConfigSpec = {
   enableTypeScript: { env: "ENABLE_BABEL_TYPESCRIPT", default: false },
+  enableDynamicImport: { env: "ENABLE_DYNAMIC_IMPORT", default: true },
```
jchip: do we want default to be `true`?

imthinhvu: It should be true, otherwise you risk breaking dynamic import for apps that already utilize it.

PR #1318, opened 2019-08-06
CEECORE-747, App can directly control webpack configs

scrajesh:


packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -1139,6 +1150,7 @@ Individual .babelrc files were generated for you in src/client and src/server
 }
 
 module.exports = function(xclap) {
+  setWebpackDevProfile();
```
jchip: There are many tasks that don't involve webpack.  You should only set this if the task ends up invoking webpack.  

And you would need the special code in `setWebpackDevProfile` to check what profile to set, because the specific task itself would know and you can set it there.

packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -142,6 +142,17 @@ function removeLogFiles() {
...
 }
 
+function checkForCustomWebpackConfig(defaultFile) {
+  const customFilePath = Path.join(process.cwd(), "webpack.config.js");
+  return Fs.existsSync(customFilePath) ? customFilePath : webpackConfig(defaultFile);
+}
+
+function setWebpackDevProfile() {
+  const devProfileTasks = ["dev", "server-admin", "wds.dev"];
+  process.env.WEBPACK_DEV_PROFILE = devProfileTasks.includes(process.argv[2] || "") ? "true" : "false";
```
jchip: It's not specific to dev only so naming it `WEBPACK_DEV_PROFILE` would be misleadning.  Name it `ELECTRODE_WEBPACK_PROFILE`

samples/hapi-app/webpack.config.js
```diff
@@ -0,0 +1,4 @@
+// Custom Webpack Config that takes control and overrides Archetype
+const { compose, env, options } = require("electrode-archetype-react-app-dev/config/webpack");
+
+module.exports = compose(options);
```
jchip: show some actual customization example.

scrajesh: Am working on simple customization examples, will update the PR once it's done.

packages/electrode-archetype-react-app/arch-clap.js
jchip: you would *not* need

scrajesh: Now, it's set at the specific task that is currently executed. But only if it's required.

packages/electrode-archetype-react-app/arch-clap.js
scrajesh: modified it.

samples/hapi-app/webpack.config.js
scrajesh: Added an example to add a plugin while on development environment.

packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -918,27 +937,39 @@ Individual .babelrc files were generated for you in src/client and src/server
...
+          `--watch --watch-aggregate-timeout 2000`,
+          archetype.webpack.enableHotModuleReload ? `--hot` : ``,
+          `--config`,
+          quote(checkForCustomWebpackConfig("webpack.config.dev.js")),
+          `--progress --colors`,
+          `--port ${archetype.webpack.devPort}`,
+          `--host ${archetype.webpack.devHostname}`
+        ],
+        {
+          xclap: { delayRunMs: 2000 },
```
jchip: why this delay?

packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -142,6 +142,17 @@ function removeLogFiles() {
   } catch (e) {} // eslint-disable-line
 }
 
+function setWebpackProfile(profile) {
+  process.env.ELECTRODE_WEBPACK_PROFILE = profile || "production";
+  logger.info(`Electrode Webpack Profile set to ${process.env.ELECTRODE_WEBPACK_PROFILE}`);
+}
+
+function checkForCustomWebpackConfig(defaultFile) {
```
jchip: shouldn't this use `util/check-custom-config.js`?

packages/electrode-archetype-react-app-dev/config/webpack/util/check-custom-config.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+const Fs = require("fs");
+const Path = require("path");
+
+function checkForCustomWebpackConfig(reqConfigFile) {
+  const customFilePath = Path.join(process.cwd(), "webpack.config.js");
+  const defaultFile = reqConfigFile || Path.join("../", "webpack.config.js");
```
jchip: returning a default relative "../webpack.config.js" that caller would `require` is not deterministic.  what if caller's dir location is different and `require("../webpack.config.js")` is invalid?  Either return absolute path or just throw if caller didn't pass reqConfigFile.

packages/electrode-archetype-react-app/arch-clap.js
scrajesh: the utils is for `electrode-archetype-react-app-dev` package.

packages/electrode-archetype-react-app/arch-clap.js
scrajesh: removed it.

packages/electrode-archetype-react-app/arch-clap.js
scrajesh: check-custom-config is for `electrode-archetype-react-app` package

packages/electrode-archetype-react-app-dev/config/webpack/util/check-custom-config.js
scrajesh: modified it. good catch, thanx!

jchip:
Our old webpack customization also look for `CWD/webpack.config.js` for user's custom code.

How can we avoid that conflict?

scrajesh:
> Our old webpack customization also look for `CWD/webpack.config.js` for user's custom code.
> 
> How can we avoid that conflict?

But is [this](https://github.com/electrode-io/electrode/tree/master/packages/electrode-archetype-react-app#extending-webpack-configuration) the way to extend webpack config as per `electrode-archetype-react-app` ?

jchip:
> But is this the way to extend webpack config as per electrode-archetype-react-app ?

PR #1352, opened 2019-09-19
make stylus optional

jchip:


packages/electrode-archetype-opt-stylus/optional-check.js
```diff
@@ -0,0 +1,188 @@
+"use strict";
```
christianlent: I'm sure this question is constantly asked, but is there anyway to DRY up the optional-checks without increasing complexity too much?

jchip: npm runs `preinstall` w/o installing dependencies so can't, but that file is maintained at a single location, and only copy to each one.  But this will not be needed with the v7 plan to remove optional-check.

PR #1353, opened 2019-09-19
make less support optional

jchip:


packages/electrode-archetype-opt-less/README.md
```diff
@@ -0,0 +1,41 @@
+# Electrode Archetype Option Stylus
```
christianlent: Are you mentioning Stylus here intentionally?

divyakarippath: Entire document refers to stylus instead of less, including the optional parameter name and package name

PR #1356, opened 2019-09-22
cleanup/refactor/fix app control webpack config 

jchip:


samples/poc-subapp/src/06.suspense-demo/subapp-suspense-demo.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import { loadSubApp } from "subapp-web";
+import PropTypes from "prop-types";
+
+let data;
+
+class Data1 extends React.Component {
+  constructor() {
+    super();
```
christianlent: Just curious, why are you overriding the constructor in these demos? Is it intentionally clearing out the construction props? If not, why not pass the props to `super`?

samples/poc-subapp/src/06.suspense-demo/subapp-suspense-demo.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  render(props) {
```
christianlent: And similarly, why does this take props?

samples/poc-subapp/src/06.suspense-demo/subapp-suspense-demo.js
jchip: good catch. :) thanks.

jchip: this change actually seem to got mixed up with another PR.  ahh... trying to juggle 5 PRs with git. :)

PR #1358, opened 2019-09-22
fix css module detection

jchip:
should merge after https://github.com/electrode-io/electrode/pull/1357

PR #1359, opened 2019-09-22
less style demo sample

jchip:
should merge after https://github.com/electrode-io/electrode/pull/1358

### DIFF
--- a/samples/hapi-app/src/client/components/demo-cookies.jsx
+++ b/samples/hapi-app/src/client/components/demo-cookies.jsx
@@ -18,6 +18,7 @@
 
 import React from "react";
 import custom from "../styles/custom.css"; // eslint-disable-line no-unused-vars
+import demo3 from "../styles/demo3.less";
 import Cookies from "electrode-cookies";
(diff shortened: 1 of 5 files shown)


samples/hapi-app/src/client/components/demo-cookies.jsx
```diff
@@ -60,7 +61,7 @@ class DemoCookies extends React.Component {
     return (
       <div>
         <h6 styleName={"custom.docs-header"}>Demo Non-Redux Universal Cookies</h6>
-        <p>Test cookie value: {cookieValue || "<NOT SET>"}</p>
+        <p styleName={"demo3.cookieTitle"}>Test cookie value: {cookieValue || "<NOT SET>"}</p>
```
christianlent: Was it your intent to make "demo3.cookieTitle" a string? If so, the braces are extraneous.

jchip: good catch, that means we have to update all the code. :D

PR #1363, opened 2019-09-24
pass initial props for non redux case

jchip:


packages/subapp-web/lib/load.js
```diff
@@ -119,15 +119,18 @@ module.exports = function setup(setupContext, token) {
       // console.log("subapp load", name, "useReactRouter", subApp.useReactRouter);
       let rrContext;
 
-      const createElement = Component => {
+      const createElement = (Component, initalProps) => {
```
christianlent: [nit] initalProps -> initialProps

PR #1373, opened 2019-09-27
CEECORE-965: Replace webpack [hash] with [contenthash]

divyakarippath:
Examples:

image

image


jchip:
👍 need more discussion on doing this for v7 only

PR #1375, opened 2019-09-27
Hapi 18 compatibility for electrode-archetype-react-app-dev

christianlent:
- Add @hapi/boom to dependencies for notFound error
- Decouple replyFile from inert
- Conditionally include inert or @hapi/inert in plugin list based on availability
- Unit tests (mocking middleware)


packages/electrode-archetype-react-app-dev/lib/dev-admin/dev-server.js
```diff
@@ -31,13 +39,18 @@ if (electrodeServer) {
...
     plugins: {
-      inert: {},
       webpackDevHapi: {
         module: "./dev-hapi.js",
         requireFromPath: __dirname
       }
     }
-  });
+  };
+  if (isInstalled("@hapi/inert")) {
```
jchip: inert is no longer required because we have our own `replyFile` implementation now.

packages/electrode-archetype-react-app-dev/lib/dev-admin/dev-hapi17.js
```diff
@@ -3,9 +3,10 @@
 /* eslint-disable no-console, no-magic-numbers */
 
 const Url = require("url");
-const Boom = require("boom");
+const Boom = require("@hapi/boom");
```
jchip: boom is only used at line 51 in this file and no where else, maybe just remove it, and create a simple 404 response.

jchip: the `replyNotFound` method doesn't seem to be used.  Let's just remove that and boom.

PR #1377, opened 2019-09-27
scan subapp dir for reducers and create HMR createStore method

jchip:
Redux store reducers hot module reload requires user's `createStore` method to detect code change and accept hot module update, but also need to call `store.replaceReducers`.
This is a fairly tricky thing to do automatically, so the approach is to assume user's reducers is named as `reducers[.js|.jsx]`, and generate a JS to import it and provide a `createStore` method for the subapp to hot accept module and replace reducers.

But if user's subapp provide its own `reduxCreateStore` method, then nothing is done.  User needs to handle HMR and `store.replaceReducers`

### DIFF
--- a/samples/poc-subapp/xclap.js
+++ b/samples/poc-subapp/xclap.js
@@ -2,7 +2,7 @@
  * Tell Electrode app archetype that you want to use ES6 syntax in your server code
  */
 
-process.env.SERVER_ES6 = true;
+// process.env.SERVER_ES6 = true;
 
 process.env.WEBPACK_DEV_MIDDLEWARE = true;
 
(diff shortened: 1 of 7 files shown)


samples/poc-subapp/xclap.js
```diff
@@ -2,7 +2,7 @@
  * Tell Electrode app archetype that you want to use ES6 syntax in your server code
  */
 
-process.env.SERVER_ES6 = true;
+// process.env.SERVER_ES6 = true;
```
christianlent: Since this is commented out, perhaps the above comment could be amended to indicate what the default value is?

jchip: this is for the sample, our generator is generating this. 
the archetype now is default to support ES6 for server so we would update this in the generator.

PR #1380, opened 2019-09-27
Ceecore 961

barbchoy:
- Verified sugarss dependency in electrode-archetype-opt-postcss
- Created a new demo ( copying Demo 2 ) but using *.styl instead of css to verify that it works.

### DIFF
--- a/samples/universal-react-node/fyn-lock.yaml
+++ b/samples/universal-react-node/fyn-lock.yaml
@@ -5259,6 +5259,7 @@ electrode-archetype-opt-postcss:
    postcss-nested: ^3.0.0
    postcss-preset-env: ^5.3.0
    postcss-scss: ^1.0.6
+   sugarss: ^1.0.1
 electrode-archetype-opt-pwa:
  _latest: 1.0.3-fynlocal_h
  _:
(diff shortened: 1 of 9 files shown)


samples/universal-react-node/fyn-lock.yaml
```diff
@@ -17693,7 +17693,7 @@ yargs:
   ^7.0.0: 7.1.0
  13.3.0:
   $: sha512-2eehun/8ALW8TLoIl7MVaRUrg+yCnenu8B4kBlRxj3GJGDKU1Og7sMXPNm1BYyM1DOJmTZ4YeN/Nwxv+8XJsUA==
-  _: 'https://registry.npmjs.org/yargs/-/yargs-13.3.0.tgz'
+  _: 'https://repository.cache.walmart.com/content/groups/npm/yargs/-/yargs-13.3.0.tgz'
```
datvong-wm: make sure your connected to registry.npmjs.org instead of npme.walmart.com when doing OSS version.

barbchoy: changed

PR #1386, opened 2019-09-30
defer loading app mode in config/archetype

jchip:


packages/electrode-archetype-react-app/config/load-dev.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+//
+// Try to set dev settings, if the dev archetype is available.
+// It may have been removed for production deployment.
+//
+function loadDev(options) {
+  const devOptions = checkTopDevArchetype(options.devArchetypeName);
+  if (devOptions) {
+    Object.assign(options, devOptions);
```
christianlent: Is it ideal that you're manipulating the passed argument directly? (rather than, for instance, Object.assign({}, options, dev.Options) or {...options, ...devOptions}? Is the options variable essentially a modules export? If so, will subsequent requires/imports of that options file receive the altered values?

jchip: I am not particularly worried about it.  This function is internal part of the whole load options op.  just broken up into files and functions to make the code more logical.

PR #1395, opened 2019-10-03
[major] new @xarc/creat-app package for npm init

jchip:


packages/create-app/.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
+  // Use IntelliSense to learn about possible attributes.
```
christianlent: Can you use comments in JSON? (Github seems to be annoyed anyway)

jchip: this is for vscode

PR #1400, opened 2019-10-05
prompt to overwrite dir

jchip:


packages/create-app/template/src/demo2/reducers.js
```diff
@@ -0,0 +1,15 @@
...
+    return {
+      value: store.value + 1
+    };
+  } else if (action.type === "DEC_NUMBER") {
+    return {
+      value: store.value - 1
+    };
+  }
+
+  return store || { value: 0 };
```
christianlent: Out of curiosity, why do you prefer this over default parameters?

jchip: This is actually copied from old sample. :)
Re: default param.
From code PoV, I like to see the default logic at the site of the return.  With default param, I have to look up to confirm the default when I see `return store`.

But from user PoV, default param is better because code hint could tell you that a param has default value.

I have no preference but I think default param makes more sense from user PoV.

PR #1401, opened 2019-10-07
load subapp bundles by their entry points and id

jchip:


packages/subapp-web/test/spec/util.spec.js
```diff
@@ -19,3 +22,27 @@ describe("loadAssetsFromStats", () => {
...
 });
+
+describe("getChunksById", function() {
+  const prodStats = JSON.parse(Fs.readFileSync(Path.join(__dirname, "../data/prod-stats.json")));
+  const devStats = JSON.parse(Fs.readFileSync(Path.join(__dirname, "../data/dev-stats.json")));
+
+  it("should handle production stats", () => {
+    const result = getChunksById(prodStats);
+
+    expect(result).to.have.property("_names_");
```
christianlent: If you expect the list of properties to grow, it might be more readable to switch to "to.have.all.keys"

PR #1407, opened 2019-10-10
Add additional directories to linter

christianlent:


packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -1073,7 +1073,7 @@ Individual .babelrc files were generated for you in src/client and src/server
           lint({
             ext: ".js,.jsx",
             config: eslintConfig(".eslintrc-react-test"),
-            targets: ["test/client"]
+            targets: ["test/client", "_test_/client", "__test__/client", "_tests_/client", "__tests__/client"]
```
jchip: do something with this https://github.com/electrode-io/electrode/blob/218c76acc03732fa2bcbf11eca45611f1bce17f6/packages/electrode-archetype-react-app/arch-clap.js#L1183 so the jest test dirs are shared?

PR #1408, opened 2019-10-10
update app template

jchip:


packages/create-app/README.md
```diff
@@ -5,11 +5,7 @@ Create a simple webapp for React/Node.js using the Electrode Platform.
 ## Usage
 
 ```bash
-mkdir my-app
-cd my-app
-npm init -y @xarc/app
-npm i
-npm run dev
+mkdir my-app && cd my-app && npm init -y @xarc/app && npm i && npm run dev
```
datvong-wm: nit: rather have broken out like before

PR #1427, opened 2019-10-26
use process.send to transfer isomorphic config from wds to app

jchip:
With dev-admin server, it's possible to setup IPC between webpack dev server and app server through `process.send`.  Use that to transfer isomorphic config from WDS to app, instead of a file on disk and relying on file watcher, causing all kinds of timing issues.

### DIFF
--- a/packages/electrode-archetype-react-app-dev/lib/dev-admin/admin-server.js
+++ b/packages/electrode-archetype-react-app-dev/lib/dev-admin/admin-server.js
@@ -1,6 +1,6 @@
 "use strict";
 
-/* eslint-disable no-console, no-process-exit, complexity */
+/* eslint-disable complexity */
 /* eslint-disable no-magic-numbers, max-len, max-statements, prefer-template */
 
 const Path = require("path");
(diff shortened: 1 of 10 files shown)


packages/electrode-archetype-react-app-dev/lib/dev-admin/admin-server.js
```diff
@@ -18,35 +20,32 @@ const PROXY_SERVER_NAME = "Electrode Dev Proxy";
...
-    await this.startDevServer();
+    await this.startWebpackDevServer();
     await this.startAppServer();
     if (DEV_PROXY_ENABLED) {
       // to debug dev proxy
       // await this.startProxyServer("--inspect-brk");
       await this.startProxyServer();
     }
-    setTimeout(() => this.showConsoleMenu(), 100);
+    setTimeout(() => this.showMenu(), 100);
```
christianlent: Why not `setTimeout(this.showMenu, 100);`?

jchip: that would lose `this`.  It'd need to be `setTimeout(this.showMenu.bind(this), 100)`

jchip: Typically I tend to avoid passing function directly into standard APIs like this, because you can't guarantee what params a function takes and what the API would pass to the function.
For example, `parseInt` takes two params and it's a thing where `["1", "2"].map(parseInt)` is not `[1, 2]`, because `parseInt` takes two params: `(str, radix)` and `map` pass two params: `(value, index)`

PR #1430, opened 2019-10-28
avoid overriding exist FORCE_COLOR env

jchip:


packages/electrode-archetype-react-app-dev/lib/dev-admin/admin-server.js
```diff
@@ -176,7 +176,7 @@ class AdminServer {
 
     const start = () => {
       const forkOpts = {
-        env: Object.assign({}, process.env, { FORCE_COLOR: true, ELECTRODE_ADMIN_SERVER: true }),
+        env: Object.assign({}, process.env, { ELECTRODE_ADMIN_SERVER: true }),
```
christianlent: FORCE_COLOR is also overridden in webpack-dll arch-clap-tasks - should it be removed there too?

PR #1437, opened 2019-10-31
no-ssr-sync flag added to reload client only without app server restart

sandip-fullstack:
Added the no-ssr-sync flag in dev task which will only build webpack for any client side code change and the change will load in UI without page refresh. It won't restart app server every time.

### DIFF
--- a/packages/electrode-archetype-react-app/arch-clap.js
+++ b/packages/electrode-archetype-react-app/arch-clap.js
@@ -846,7 +846,7 @@ Individual .babelrc files were generated for you in src/client and src/server
       dep: [".init-bundle.valid.log"],
       desc: "Start app's node server in watch mode with nodemon",
       task: function() {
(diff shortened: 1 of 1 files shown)


packages/electrode-archetype-react-app/arch-clap.js
```diff
@@ -867,9 +867,10 @@ Individual .babelrc files were generated for you in src/client and src/server
           nodeRunApp = quote(Path.relative(process.cwd(), serverRun));
         }
 
+        let taskArguments = this.argv.includes("--no-ssr-sync") ? "" : taskArgs(this.argv).join(" ");
```
jchip: please do `this.argv.filter( x => x !== "--no-ssr-sync" ).join( " " )`

sandip-fullstack: @jchip Please check now.

jchip (approved):

PR #1439, opened 2019-11-01
Features task

christianlent:
This creates a new archetype feature task (to use, run `clap features`).

- For each feature, show the following information: name, enabled, current version, latest version, and description
- Use `npm view` to determine the most recent available version of the feature.
- Use legacy opt eval (identical to the `optional-check.js` code) to determine whether the feature is enabled
- Order features by name
- Display warning if multiple mutually-exclusive options are selected
- Display warning if a package is enabled by but installed
- No new dependencies


packages/electrode-archetype-react-app/lib/features.js
```diff
@@ -9,6 +9,7 @@ const os = require("os");
 const exec = util.promisify(require("child_process").exec);
 const appArchetypeConfig = optionalRequire("electrode-archetype-react-app/config/archetype") || {};
 const prompts = require("prompts");
+const request = require("request");
```
jchip: `request` is dep in `electrode-archetype-react-app-dev`.  we use `appArchetypeConfig.devRequire` to ensure start looking from there.

packages/electrode-archetype-react-app/lib/features.js
```diff
@@ -62,27 +40,38 @@ class Feature {
...
-    const values = stdout
-      .replace(/[\n\r]/g, "") // No newlines
-      .match(regex)
-      .splice(1)
-      .map(parseRelaxedJson);
-    const fieldMap = {};
-    fields.forEach((field, index) => (fieldMap[field] = values[index]));
-    return fieldMap;
+  async getNpmAttributes() {
+    const pathName = `.etmp/${this.packageName}.registry.json`;
```
jchip: 1. The Electrode Tmp dir is available from `appArchetypeConfig.eTmpDir`
2. It's probably safer to do a `Path.resolve` to get the full path.

PR #1441, opened 2019-11-02
style subapps for create-app sample

jchip:


packages/create-app/template/src/demo2/subapp-demo2.js
```diff
@@ -20,9 +20,21 @@ const Demo2 = props => {
 
   return (
     <div>
-      Redux State Demo: <button onClick={() => dispatch(decNumber())}>&#8810;</button>
-      &nbsp;{value}&nbsp;
-      <button onClick={() => dispatch(incNumber())}>&#8811;</button>
+      <div
+        style={{
+          padding: "5px",
```
christianlent: Sometimes I prefer the shorthand of just using the number, e.g. `padding: 5`
For many size-based style attributes, it will assume "px" by default.

jchip: 👍 I imagine real complex apps wouldn't use styles like this.  This is a simple sample and I wanted to make it explicit.

PR #1443, opened 2019-11-03
Handle load subapp error

jchip:


packages/subapp-web/lib/load.js
```diff
@@ -51,7 +51,7 @@ module.exports = function setup(setupContext, token) {
   //
   const retrieveDevServerBundle = async () => {
     return new Promise((resolve, reject) => {
-      request(`${bundleBase}${bundleAsset.name}`, (err, resp, body) => {
+      fetch(`${bundleBase}${bundleAsset.name}`, (err, resp, body) => {
```
christianlent: I see why you did this (you're using `request` below and you don't want to shadow). However, I would find this thoroughly confusing since the fetch API is fairly popular now and it has a different call signature (particularly with its use of promises). Is there any way you could rename this to something that collides with less popular packages? Like `httpRequest`, `callbackRequest`, or something?

packages/subapp-web/lib/load.js
```diff
@@ -200,16 +200,12 @@ module.exports = function setup(setupContext, token) {
         }
         return React.createElement(
           AppContext.Provider,
-          { value: { isSsr: true, subApp, ssr: { request: req } } },
+          { value: { isSsr: true, subApp, ssr: { request: request } } },
```
christianlent: Shorthand?

PR #1450, opened 2019-11-15
If devOnly is undefined, default to true.

christianlent:
Also, search for devOnly in locally installed package if not available in npm registry data.

### DIFF
--- a/packages/electrode-archetype-react-app/lib/features.js
+++ b/packages/electrode-archetype-react-app/lib/features.js
@@ -218,10 +218,18 @@ class Feature {
     }
     const dependencies = { ...pkg.dependencies };
     const devDependencies = { ...pkg.devDependencies };
-    if (enabled && this.electrodeOptArchetype.devOnly) {
+
+    let devOnly = true;
+    if (this.electrodeOptArchetype.devOnly !== undefined) {
+      devOnly = this.electrodeOptArchetype.devOnly;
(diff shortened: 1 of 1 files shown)


packages/electrode-archetype-react-app/lib/features.js
```diff
@@ -218,10 +218,18 @@ class Feature {
     }
     const dependencies = { ...pkg.dependencies };
     const devDependencies = { ...pkg.devDependencies };
-    if (enabled && this.electrodeOptArchetype.devOnly) {
+
+    let devOnly = true;
+    if (this.electrodeOptArchetype.devOnly !== undefined) {
+      devOnly = this.electrodeOptArchetype.devOnly;
+    } else if (this.package && this.package.electrodeOptArchetype.devOnly !== undefined) {
+      devOnly = this.package.electrodeOptArchetype.devOnly;
```
datvong-wm: what's `this.package` vs `this._package`?

datvong-wm: oh got it. nvm

christianlent: `_package` is the actual object containing the `package.json` contents (really, a cache if you will), and `package` is a getter that is capable of fetching the `package.json` contents and caching it to _package. No consumer of this class should access _package directly.

PR #1453, opened 2019-11-15
Ensure "prompts" is available for features clap task

christianlent:


jchip:
this should be in dev archetype.  use `devRequire` to require it.

christianlent:
That's what I would have thought! But, when I use the create app task that you mentioned in the defect, I don't think it installed prompts at all. Here is the line where I include prompts: what am I doing wrong?
https://github.com/christianlent/electrode/blob/dbf15c52f7972209e4cc470ee420f749cb7b835a/packages/electrode-archetype-react-app/lib/features.js#L17

PR #1466, opened 2019-12-11
Should detect all Unhandled rejection messages as an error in log-parser

christianlent:


jchip:
all `Unhandled rejection` should be treated as error.  The PropTypes one was just an example.

PR #1474, opened 2019-12-19
Ce1134

datvong-wm:


packages/electrode-archetype-react-app-dev/lib/webpack-dev-fastify.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+/* eslint-disable no-console, no-magic-numbers */
+
+const fastifyPlugin = require("fastify-plugin");
```
jchip: need to add this to `dependencies`

PR #1493, opened 2020-01-13
update create-app to use archetype ver7

jchip:


packages/create-app/template/src/demo2/subapp-demo2.js
```diff
@@ -42,10 +42,7 @@ const Demo2 = props => {
 const mapStateToProps = state => state;
 
 export default reduxLoadSubApp({
-  Component: connect(
-    mapStateToProps,
-    dispatch => ({ dispatch })
-  )(Demo2),
+  Component: connect(mapStateToProps, dispatch => ({ dispatch }))(Demo2),
```
christianlent: This is totally fine, but it's a little hard read (at least to my eyes), and I would imagine that a lot of people will want to read this to understand how the boilerplate code works.

jchip: could you rewrite it please? :)

PR #1497, opened 2020-01-17
First pass of an Electrode subapp guide

christianlent:
There are no inbound links to this page yet.

### DIFF
--- a/docs/guides/subapps.md
+++ b/docs/guides/subapps.md
@@ -1,0 +1,279 @@
+# Electrode Subapps
+
+Electrode Subapps is a mechanism for code splitting within a larger application. There are several advantages to creating a web application using subapps:
+
+- Superior code isolation is conducive to development across multiple teams
+- Smaller initial payloads loads through lazy-loadable subapps
+- Better application scalability
+
+## Getting Started
+
+To easily start using the npm initializer, run the following command:
+
+```bash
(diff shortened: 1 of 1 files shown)


docs/guides/subapps.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Getting Started
+
+To easily start using the npm initializer, run the following command:
+
+```bash
+mkdir my-app && cd my-app && npm init -y @xarc/app && npm i && npm run dev
+```
+
+This will create, install, and run an Electrode web application with a few simple subapps.
```
dmikey: Personally I like when this break down, highlights the command that was used above, so the engineer has context of how the above line was composed.

docs/guides/subapps.md
```diff
@@ -0,0 +1,279 @@
...
+
+## What is a subapp?
+
+- What is a subapp?
+
+  At its core, a subapp is simply a display component. Currently, this can be either or React or Preact component, but more UI libraries may be supported in the future.
```
dmikey: one too many ors? React, Preact?

docs/guides/subapps.md
christianlent: I do like that! Would you separate into multiple lines and add comments to each line?

PR #1532, opened 2020-02-13
New pbundle option "packReduxData" for SSR redux state

christianlent:


jchip:
it should be better to do this in `realizeReduxStore`

jchip:
also probably should do something related to this https://github.com/electrode-io/electrode/blob/1529a7e99cf172ff915477a30a414d054d66edbf/packages/subapp-pbundle/lib/framework-lib.js#L144

- do that if no `packReduxData` exist
- `packReduxData` part should check `attachInitialState ` flag

packages/subapp-pbundle/lib/framework-lib.js
```diff
@@ -169,8 +170,16 @@ class FrameworkLib {
...
+      const { subApp, subAppServer } = this.ref;
       this.store = this.store.realize();
       await this.signalStoreReady();
+
+      const packReduxData = subAppServer.packReduxData || subApp.packReduxData;
+      // if subapp didn't request to skip sending initial state and packReduxData was specified,
+      //  then stringify packReduxData's return value and attach it to the index html.
+      if (packReduxData && subAppServer.attachInitialState !== false) {
+        this.initialStateStr = JSON.stringify(packReduxData(this.store));
+      }
```
jchip: this should be done outside of the if, because a store may have been realized already.

Also, only do this if `this.initialStateStr` is falsy

PR #1630, opened 2020-04-28
added: mechanism to defer initial state read for sub apps

HenrikJoreteg:
hey @jchip and @christianlent! I can't actually seem to get fyn bootstrap to run successfully, so i'm struggling to run tests, etc. 

So i figured I'd at least send this PR to clarify what I'm suggesting we do. 

Basically, this would allow us to run `instance.getInitialState()` and thereby defer the reading of that data instead of blocking on that with the inline script (as we discussed).


HenrikJoreteg:
there's a really good chance this breaks other things 😅 . I'm mainly proposing an idea here, not necessarily saying this is merge-worthy.

PR #1641, opened 2020-05-12
Must add babel-plugin-react-css-modules as a direct dependency

datvong-wm:
According to installation instructions

https://github.com/gajus/babel-plugin-react-css-modules#installation
> you must install babel-plugin-react-css-modules as a direct dependency of the project

It is still a part of electrode-archetype-opt-postcss as an optional dependency


datvong-wm:
Close and cancel PR.  Some people don't want to add this dependency as it doesn't affect all users..  Will ask teams to manually update their project.

PR #1642, opened 2020-05-15
[minor]: add fastify support to subapp-server

jchip:


samples/poc-subapp-min-fastify/config/default.js
```diff
@@ -0,0 +1,32 @@
...
+  return x !== null && !isNaN(x) ? x : defaultListenPort;
+};
+
+module.exports = {
+  plugins: {
+    "webpack-dev": {
+      module: "@xarc/app-dev/lib/webpack-dev-fastify",
+      enable: process.env.WEBPACK_DEV === "true"
+    },
+    "subapp-server": { options: { insertTokenIds: true } }
```
divyakarippath, line 17: Since we are not using universal hapi plugin, I am thinking if we need to directly include fastify plugin here, like `subapp-server/lib/fastify-plugin/`.

jchip: subapp-server exports `{ hapiPlugin, fastifyPlugin }`, and electrode-server or fastify-server will pick the one they can handle. :)

PR #1643, opened 2020-05-18
feat/extensible-xarc

cdaringe:
# problem

xarc archetype tasks are not configurable onload

# solution

- rather than induce side-effects on-require, instead, take _most of the effects_ and only execute the effects on function call. in other words, take _everything to do with the archetype_ and lazily execute the effects inside a function.  now, what was once an effectful import, is now a lazy, and thus can be configured.

e.g.

```ts
const archetype = require('<archetype-pkg>/config/archetype')({
  flagA: true,
  flagB: false,
  retries: 100
})
```

vs

```ts
process.env.IS_FLAG_A_ENABLED = '1'; // side-effect
require('<archetype-pkg>/config/archetype'); // more implicit side effects
```

the totality of the PR could be summarized by taking existing code that looks like this:

```js
require('x') // effectful
const y = require('z');
y.doStuff()  // effectful
module.exports = { y }
```

into something like:

```js
module.exports = (opts) => {
  require('x') // effectful, but at least user flow controlled
  const y = require('z');
  y.doStuff(opts.y)  // effectful, but at least user flow controlled
}
```

# discussion

**this changeset is massive**, but can be trimmed down extensively by reviewing ignoring whitespace: https://github.com/electrode-io/electrode/pull/1643/files?w=1
 
on `require('@xarc/app[-dev]')`, a variety of side-effects and eager binding occur. dynamic requires to various configuration files are eagerly evaluated, and the ability to _tune_ the resultant tasks and configuration is voided _unless_ ENV vars are set in advanced before the require statement is executed. a major short-coming of this effort is that in many places, configuration _only happens via the ENV_, specifically in development mode when firing off in child processes. for now, i ask that we disregard that issue, as there are strategies to mitigate that shortcoming, such as modifiying the child process env, writing temp config, sending config over ipc, launching with additional flags, and more.

this PR attempts to:

- not change the core logic/flows of existing archetypes, beyond the archetype exports
- move all effects to inside functions
- _improve_ some amount of backward compatibility by still having the archetypes be singletons. to not do so would be very challenging given the way how the archetypes are require'd from so many sources and mutated by _many parties_


cdaringe:
hmm. travis tests are ok on node 10, but gh actions are not 🤔

cdaringe:
closing in favor of tidier git history in #1645

PR #1656, opened 2020-05-20
 inline js loading in webapp init

yishengjiang99:


yishengjiang99:
loading runtime bundle inline with html in subapp init.

yishengjiang99:
inlineload-runtimejs
subappload

see screenshots of source files being generated with the runtime.js loaded inline.

PR #1659, opened 2020-05-26
Preload scripts and load styles in <head>

divyakarippath:


packages/subapp-web/lib/load.js
```diff
@@ -210,6 +219,8 @@ Response: ${err || body}`
 
       const { request } = context.user;
 
+      context.user.numOfSubapps = context.user.numOfSubapps || 0;
```
jchip, line 222: if this initializes it to `0`, wouldn't that cause line 326 below to change it to `-1`?

jchip:
could you add some tests please?

packages/subapp-web/lib/load.js
divyakarippath: Before that processSubapp() (line 348) will be called where the count will be incremented first. But still, I think its good to add an additional check before decrementing the count. I will add that.

PR #1661, opened 2020-05-27
Inline js runtime bundle

yishengjiang99:
screenshot-pr-1661


packages/subapp-web/lib/init.js
```diff
@@ -34,10 +30,18 @@ module.exports = function setup(setupContext) {
     basePath: ""
   };
 
+  const approot = process.env.APP_ROOT_DIR || Path.resolve() || process.cwd();
```
jchip, line 33: - don't use env
- `Path.resolve("dist")` should do

packages/subapp-web/lib/init.js
```diff
@@ -34,10 +30,18 @@ module.exports = function setup(setupContext) {
...
+  let inlineRuntimeJS = "";
+  const runtimeFSPath = cdnJsBundles.runtime && Path.join(approot, "dist", cdnJsBundles.runtime);
+  if (runtimeFSPath && Fs.existsSync(runtimeFSPath)) {
```
jchip, line 36: only do this for `NODE_ENV` === `production`

packages/subapp-web/test/data/mock-app/dist/stats-with-runtime.json
```diff
@@ -0,0 +1,107 @@
+{
```
jchip, line 1: you need to use stats from production build.  get it from `dist/server` by running `clap build`

jchip (changes requested):

yishengjiang99:
screenshot-pr-1661

PR #1669, opened 2020-06-09
dep: xaa 1.6.0

jchip:


packages/subapp-web/lib/start.js
```diff
@@ -24,26 +25,20 @@ module.exports = function setup() {
       /*
        */
       const waitForPrepare = async (info, mapCtx) => {
-        if (mapCtx.failed) {
-          return;
-        }
+        mapCtx.assertNoFailure();
```
datvong-wm, line 28: try - catch?

jchip: how? details?

PR #1672, opened 2020-06-12
Web SSR reporting

divyakarippath:


packages/subapp-web/lib/init.js
```diff
@@ -16,6 +16,8 @@ module.exports = function setup(setupContext) {
   // TODO: in webpack dev mode, we need to reload stats after there's a change
   //
 
+  const metricReport = setupContext.routeOptions.reporting || {};
```
jchip, line 19: use `lodash.get`

packages/subapp-web/lib/init.js
```diff
@@ -68,6 +70,13 @@ ${inlineRuntimeJS}
       runtimeEntryPoints.forEach(ep => {
         context.user.includedBundles[ep] = true;
       });
+
+      if (metricReport.enable && metricReport.reporter) {
+        context.user.emitter = util.getEventEmiiter(metricReport.reporter);
+      } else {
+        // eslint-disable-next-line
+        console.warn(`Warning: Metric reporting for ssr not enabled or no reporter specified.`);
```
jchip, line 78: isn't this going to log for every request?

packages/subapp-web/lib/init.js
```diff
@@ -68,6 +70,13 @@ ${inlineRuntimeJS}
       runtimeEntryPoints.forEach(ep => {
         context.user.includedBundles[ep] = true;
       });
+
+      if (metricReport.enable && metricReport.reporter) {
+        context.user.emitter = util.getEventEmiiter(metricReport.reporter);
```
jchip, line 75: name this to something more specific?

idea: `xarcSSREmitter`

packages/subapp-web/lib/init.js
divyakarippath: oh yes, you are right. I should probably move this to **subapp-server**. Was trying to avoid changes to another package.

packages/subapp-web/lib/start.js
```diff
@@ -85,6 +93,8 @@ module.exports = function setup() {
             if (group !== "_") {
               mapCtx.assertNoFailure();
 
+              startTimes[`load-grp-${group}`] = Date.now();
```
jchip, line 96: if `startTimes` is only used within this closure, then it can simply be a variable.

packages/subapp-web/lib/start.js
```diff
@@ -63,6 +64,7 @@ module.exports = function setup() {
 
       // default group _ subapps should all run independently
       if (xarcSubappSSR._) {
+        startTimes[`load-subapps`] = Date.now();
```
jchip, line 67: if `startTimes` is only used within this closure, then it can simply be a variable.

packages/subapp-web/lib/start.js
```diff
@@ -108,6 +127,20 @@ module.exports = function setup() {
...
                 { concurrency }
               );
+              if (xarcSSREmitter) {
+                xarcSSREmitter.emit("web_ssr", {
+                  action: `render-group`,
+                  group,
+                  labels: ["group"],
+                  duration: Date.now() - startTimes[`render-grp-${group}`]
+                });
+                xarcSSREmitter.emit("web_ssr", {
```
jchip, line 137: what's the context behind emitting two events with the same time?  please add comment to explain that.

Please use the same now timestamp for both.

divyakarippath: `render-group` event is to capture the duration for just rendering of groups, whereas `load-group` captures the total processing time for that group. I am capturing the start times of the events accordingly. I was a little skeptical about the `load-group` event. But thought, its nice to have one events for the total time. Please let me know your suggestions here.

jchip: so `load-group` is the total time of prepare + render?
that's fine, but maybe a better name?  

name suggestions:

`group-total-${group}`
`group-render-${group}`

PR #1700, opened 2020-07-14
Loading subapps from a different repo

divyakarippath:
Directory structure of individual subapps:

```
my-sub-app
├── README.md
├── package.json
├── src
........├── index.js
........├── subapp.js
........├── server.js
........├── reducers.js
```
Each subapp should be exporting the manifest and sub-app options as below

```
const subAppOptions = {
  serverSideRendering: false
};

const manifest = {
  type: "app",
  name: "Demo1",
  entry: require.resolve("./subapp"),
  serverEntry: require.resolve("./server"),
  reducers: require.resolve("./reducers")
};

module.exports = { subAppOptions, manifest `};
```

The users must provide absolute paths for `entry`, `serverEntry` and `reducers`
**Publishing** : The code must be transpiled to node executable format before publishing the package.

**Main app:**

Create a file named `subapps-modules.js` in `src` and export the manifests for all the subapp modules

```
const demo1 = require("demo1");
const demo2 = require("demo2");

module.exports = {
  demo1: demo1.manifest,
  demo2: demo2.manifest
};
```
Update the routes file to load the subapp module.

```
export default {
"/home": {
    pageTitle: "Welcome to home",
    subApps: ["demo1", "demo2"]
  }
}
```

Main app can also customize the options for subapp as below.

```
"/home": {
    pageTitle: "Welcome to home",
    subApps: [["demo1", subAppOptions], ["demo2", subAppOptions]]
  }
}
```

### DIFF
--- a/packages/subapp-server/lib/setup-hapi-routes.js
+++ b/packages/subapp-server/lib/setup-hapi-routes.js
@@ -115,8 +115,7 @@ function setupRouteRender({ subAppsByPath, srcDir, routeOptions }) {
       // else: assume dir under srcDir
       // TBD: handle it being a module
       if (x.indexOf("/") === -1) {
-        const xSrcDir = options.srcDir || "lib";
-        x = Path.resolve("node_modules", x, xSrcDir);
+        x = Path.dirname(require.resolve(`${x}`));
       }
       return {
         subapp: subAppsByPath[Path.isAbsolute(x) ? x : Path.resolve(srcDir, x)],
--- a/packages/subapp-server/lib/setup-hapi-routes.js
+++ b/packages/subapp-server/lib/setup-hapi-routes.js
@@ -112,9 +112,9 @@ function setupRouteRender({ subAppsByPath, srcDir, routeOptions }) {
         x = x[0];
       }
       // absolute: use as path
+      // is module: resolve module path
       // else: assume dir under srcDir
-      // TBD: handle it being a module
-      if (x.indexOf("/") === -1) {
+      if (options.module) {
         x = Path.dirname(require.resolve(`${x}/package.json`));
       }
       return {
--- a/packages/subapp-server/lib/setup-hapi-routes.js
+++ b/packages/subapp-server/lib/setup-hapi-routes.js
@@ -108,12 +108,16 @@ function setupRouteRender({ subAppsByPath, srcDir, routeOptions }) {
     routeOptions.__internals.subApps = [].concat(routeOptions.subApps).map(x => {
       let options = {};
       if (Array.isArray(x)) {
-        options = x[1];
+        options = x[1] || {};
         x = x[0];
       }
       // absolute: use as path
       // else: assume dir under srcDir
       // TBD: handle it being a module
+      if (x.indexOf("/") === -1) {
+        const xSrcDir = options.srcDir || "lib";
+        x = Path.resolve("node_modules", x, xSrcDir);
+      }
       return {
         subapp: subAppsByPath[Path.isAbsolute(x) ? x : Path.resolve(srcDir, x)],
         options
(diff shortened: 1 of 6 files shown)


packages/subapp-server/lib/setup-hapi-routes.js
```diff
@@ -108,12 +108,15 @@ function setupRouteRender({ subAppsByPath, srcDir, routeOptions }) {
...
       let options = {};
       if (Array.isArray(x)) {
-        options = x[1];
+        options = x[1] || {};
         x = x[0];
       }
       // absolute: use as path
       // else: assume dir under srcDir
       // TBD: handle it being a module
+      if (x.indexOf("/") === -1) {
```
jchip, line 117: what's this trying to do?

packages/subapp-util/lib/index.js
```diff
@@ -29,6 +30,15 @@ function removeExt(f) {
   return ix > 0 ? f.substring(0, ix) : f;
 }
 
+const validateModuleManifest = manifest => {
+  const entries = _.pick(manifest, ["entry", "serverEntry", "reducers"]);
+  Object.keys(entries).forEach(x => {
+    if (!Path.isAbsolute(entries[x])) {
+      throw new Error(`Error: manifest ${x} requires absolute path`);
```
jchip, line 37: error message is not very clear

packages/subapp-server/lib/setup-hapi-routes.js
divyakarippath: This is to handle the scenario where subapp name is just the module name (when loading from a separate repo)

```
"/demo1": {
    pageTitle: "Welcome to Demo1",
    subApps: [["demo1", demo1.subAppOptions]],
  },
```
For the sub-apps from a separate repo, its path is set as  ```Path.dirname(require.resolve(`${modName}/package.json`))``` in `subAppsByPath`. Same logic is used here to retrive the details from the map.

Another option I was considering it is to set `module:true` in `subappOptions` and use that flag instead of checking for `/` in url. But that would be one additional thing to be set by the main app.

In existing code, if path is not absolute, we assume it's relative to app src dir.

divyakarippath: Added `module:true` flag to replace `(x.indexOf("/") === -1)` to handle module name being `@scope/name` also.

packages/subapp-util/lib/index.js
divyakarippath: Updated the error message in latest commit

packages/subapp-server/lib/setup-hapi-routes.js
divyakarippath: Updated the code to remove `module:true` flag

packages/subapp-server/lib/setup-hapi-routes.js
```diff
@@ -106,17 +106,25 @@ function setupRouteRender({ subAppsByPath, srcDir, routeOptions }) {
...
       // absolute: use as path
+      // module: resolve module path
       // else: assume dir under srcDir
-      // TBD: handle it being a module
+      if (!x.startsWith(".") && !x.startsWith("/")) {
+        const subAppPath = optionalRequire.resolve(x);
+        if (subAppPath) {
+          const { manifest, subAppOptions } = require(x);
+          x = manifest ? Path.dirname(subAppPath) : x;
+          options = options || subAppOptions;
```
divyakarippath, line 122: Added this to support passing subapp options from main app also. This will be useful in use-cases where main app wants to customize the options for individual subapps.

samples/react-jest-app/package.json
```diff
@@ -62,7 +62,7 @@
     "@xarc/app-dev": "^5.3.4",
     "electrode-archetype-opt-critical-css": "^1.0.3",
     "electrode-archetype-opt-eslint": "^1.0.3",
-    "electrode-archetype-opt-jest": "^25.0.0",
+    "electrode-archetype-opt-jest": "^26.0.0",
```
divyakarippath, line 65: Changed to match the version with latest publish

samples/demo-tree-shaking/package.json
```diff
@@ -55,6 +55,7 @@
     "lodash": "^4.17.10",
     "milligram": "^1.3.0",
     "react-intl": "^3.11.0",
+    "intl-format-cache": "4.2.46",
```
divyakarippath, line 58: To fix the build issue

PR #1708, opened 2020-07-23
   convert xarc-app to ts

jchip:


packages/xarc-app/src/index.ts
```diff
@@ -0,0 +1,207 @@
...
+         * Set this to false to avoid loading the ignore hook.
+         */
+        ignoreStyles?: boolean;
+        /**
+         * Set to false to avoid loading node.js require hook to handle isomorphic assets
+         */
+        isomorphicExtendRequire?: boolean;
+      }
+    | Function,
+  callback?: Function
```
cdaringe, line 114: unsure who may use this, but do we have any extra info on this callback? maybe it's actually `(config: Config) => void`? not saying it is, just curious though!

packages/xarc-app/src/index.ts
```diff
@@ -0,0 +1,207 @@
...
+    cssModuleHook(opts);
+  } else if (options.ignoreStyles !== false) {
+    optionalRequire("ignore-styles");
+  }
+
+  if (options.isomorphicExtendRequire !== false) {
+    return isomorphicExtendRequire();
+  }
+
+  return callback ? callback() : Promise.resolve();
```
cdaringe, line 191: hmmm, my last comment didn't go through on this line, but i answer my question above :).

perhaps instead of `Function` we use `() => Promise<void>` for the type for `callback`?

jchip: yup, definitely will be reiterating to improve types.

PR #1716, opened 2020-08-05
Load cdn mapping script for non js/css assets only if cdn option isenabled

divyakarippath:
https://jira.walmart.com/browse/CEECORE-1751

### DIFF
--- a/packages/subapp-web/lib/init.js
+++ b/packages/subapp-web/lib/init.js
@@ -10,9 +10,12 @@ const _ = require("lodash");
 const assert = require("assert");
 
 module.exports = function setup(setupContext) {
+  const cdnEnabled = setupContext.routeOptions.cdn && setupContext.routeOptions.cdn.enable === true;
   const distDir = process.env.NODE_ENV === "production" ? "../dist/min" : "../dist/dev";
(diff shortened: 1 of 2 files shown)


packages/subapp-web/src/subapp-web.js
```diff
@@ -1,6 +1,10 @@
 // xarc subapp client side lib version 1
 // load into window.xarcV1 as a global
 (function (w) {
+  if (!w._wml) {
```
divyakarippath, line 4: @jchip we were already initializing `w._wml` in `cdn-map.js` (https://github.com/electrode-io/electrode/blob/master/packages/subapp-web/src/cdn-map.js#L31). 

But now, since we don't load `cdn-map.js` for dev, added this here also. 

If we dont initialize this, it will throw error for `electrode-ui-config`  where we try to set config values in `window._wml`

packages/subapp-web/lib/init.js
```diff
@@ -10,9 +10,12 @@ const _ = require("lodash");
 const assert = require("assert");
 
 module.exports = function setup(setupContext) {
+  const cdnEnabled = setupContext.routeOptions.cdn && setupContext.routeOptions.cdn.enable === true;
```
jchip, line 13: lodash, `_.get`

PR #1717, opened 2020-08-07
Refactoring to improve performance

divyakarippath:
https://jira.walmart.com/browse/CEECORE-1748

### DIFF
--- a/packages/subapp-util/lib/index.js
+++ b/packages/subapp-util/lib/index.js
@@ -285,7 +285,7 @@ function loadSubAppByName(name) {
   return container[name];
 }
 
-function loadSubAppServerByName(name) {
+function loadSubAppServerByName(name, serverSideRendering) {
   const manifest = subAppManifest()[name];
   const { subAppDir, serverEntry } = manifest;
 
@@ -295,16 +295,16 @@ function loadSubAppServerByName(name) {
     );
   } else if (serverEntry === false) {
     return {};
+  } else if (serverSideRendering) {
+    const subapp = es6Require(
+      manifest.module ? manifest.entry : Path.resolve(appSrcDir(), subAppDir, manifest.entry)
+    );
+
+    return {
+      StartComponent: subapp.Component
+    };
   }
-
-  // generate a server from subapp's main file
-  const subapp = es6Require(
-    manifest.module ? manifest.entry : Path.resolve(appSrcDir(), subAppDir, manifest.entry)
-  );
-
-  return {
-    StartComponent: subapp.Component
-  };
+  return {};
 }
 
 function refreshSubAppByName(name) {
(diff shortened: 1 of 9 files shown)


packages/subapp-util/lib/index.js
```diff
@@ -295,16 +295,16 @@ function loadSubAppServerByName(name) {
     );
   } else if (serverEntry === false) {
     return {};
-  }
-
-  // generate a server from subapp's main file
```
divyakarippath, line 300: Modified to generate server from subapps main file only if ssr is enabled

packages/subapp-web/lib/init.js
```diff
@@ -62,8 +62,11 @@ ${inlineRuntimeJS}
 </script>`;
 
   // check if any subapp has server side code with initialize method and load them
-  const subAppServers = Object.keys(subappUtil.getAllSubAppManifest())
```
divyakarippath, line 65: Runthis only for subapps for that specific route. In `subapp-server`, we set the subapps for each route to `setupContext.routeOptions.__internals`

packages/subapp-web/lib/load.js
```diff
@@ -193,10 +193,9 @@ Response: ${err || body}`
 
   const loadSubApp = () => {
     subApp = loadSubAppByName(name);
-    subAppServer = loadSubAppServerByName(name);
+    subAppServer = loadSubAppServerByName(name, true);
```
divyakarippath, line 196: Now we will call `loadSubapp` only if ssr is enabled. Setting second argument as `true` so that `loadSubAppServerByName` will generate server from subapp's main file if `serverEntry` is not there.

packages/subapp-web/lib/load.js
```diff
@@ -193,10 +193,9 @@ Response: ${err || body}`
...
   };
 
-  loadSubApp();
```
divyakarippath, line 199: Need this only if ssr is enabled

packages/subapp-web/lib/load.js
```diff
@@ -221,10 +220,6 @@ Response: ${err || body}`
 
       context.user.numOfSubapps = context.user.numOfSubapps || 0;
 
-      if (request.app.webpackDev && subAppLoadTime < request.app.webpackDev.compileTime) {
```
divyakarippath, line 224: Needto call `loadSubApp` only if ssr is enabled

packages/subapp-web/lib/load.js
```diff
@@ -337,14 +332,6 @@ ${stack}`,
       };
 
       const processSubapp = async () => {
-        const ref = {
```
divyakarippath, line 340: Need this only for ssr flow. So moved it to line 358

PR #1726, opened 2020-08-27
fix: handle windows \ path separator

jchip:


packages/electrode-node-resolver/lib/node-resolver.js
```diff
@@ -63,8 +63,10 @@ function resolve(req, atPath) {
 
   if (!resolved) return undefined;
 
-  const ix = resolved.lastIndexOf(name);
-  const path = resolved.substr(0, ix + name.length);
+  // ensure windows \ path separator changed to / when matching for require
+  // req name, which should always be using /.
+  const ix = resolved.replace(/\\/g, "/").lastIndexOf(`${name}/`);
```
datvong-wm, line 68: added an extra "/" to the name

jchip: yes, intended, to ensure matching the right name, in case a module exports a file with the name of itself.

PR #1745, opened 2020-10-05
feat: add content hash to bundle names

divyakarippath:


jchip:
please double check: must be done in production mode only, otherwise it will break hot module reload in development mode (because the bundle is different every time due to the hash).

divyakarippath:
> please double check: must be done in production mode only, otherwise it will break hot module reload in development mode (because the bundle is different every time due to the hash).
For development mode, we are overriding this in `dev` partial. 

```
output: {
      publicPath: makePublicPath(),
      filename: "[name].bundle.dev.js",
      chunkFilename: "[name].bundle.dev.js"
    },
```

But to be safe, I can create another partial for `prod` and set it there.

PR #1760, opened 2020-11-19
[Minor] [Feature] added xarc path as cwd in packages

durrab:


packages/xarc-app-dev/src/config/archetype.ts
```diff
@@ -15,8 +15,20 @@ let cachedArchetype = null;
...
  */
 module.exports = function getDevOptions(user: XarcOptions = {}) {
+  //checking for cwd from xclap or from env
+  const cwd = user.cwd || process.env.XARC_CWD || process.cwd();
+  process.env.XARC_CWD = cwd;
+
   if (cachedArchetype) {
+    // if cached is already runnig
     cachedArchetype._fromCache = true;
+    if (
```
jchip, line 25: explain this

packages/xarc-app-dev/src/config/archetype.ts
```diff
@@ -25,7 +37,7 @@ module.exports = function getDevOptions(user: XarcOptions = {}) {
   const legacy = getDevArchetypeLegacy();
   // try to read xarc-options.json if it exist and merge it into legacy
   const xarcOptions = loadXarcOptions();
-  user = _.merge({}, xarcOptions, user);
+  user = _.merge({}, user, xarcOptions);
```
jchip, line 40: user should have priority over `xarcOptions`

packages/xarc-app-dev/src/config/archetype.ts
```diff
@@ -40,11 +52,17 @@ module.exports = function getDevOptions(user: XarcOptions = {}) {
...
   _.merge(legacy, {
     ...user,
+    cwd,
     webpackOptions: undefined,
     babelOptions: undefined,
     addOnFeatures: undefined
   });
 
+  //added XARC CMD option
+  if (!_.isNil(user) && !_.isNil(user.cwd)) {
```
jchip, line 62: why do you need it in `legacy.options.cwd` in addition to `legacy.cwd`?

packages/xarc-app-dev/src/config/opt2/xarc-options.ts
```diff
@@ -76,4 +76,9 @@ export type XarcOptions = {
    * options related to babel transpiler
    */
   babelOptions?: BabelOptions;
+
+  /**
+   * option for app cwd to be used from xarc-options
```
jchip, line 81: what does "to be used from xarc-options" mean?

packages/xarc-app/src/index.ts
```diff
@@ -297,7 +299,7 @@ if (!getAppMode().hasEnv()) {
   const guessAppSrcDir = () => {
     if (module.parent && module.parent.filename) {
       const fn = module.parent.filename;
-      const dir = fn.substr(process.cwd().length + 1).split("/")[0];
+      const dir = fn.substr(detectXARCPath(archetype.options.cwd)).split("/")[0];
```
jchip, line 302: the new code is wrong, look at the old code.

packages/xarc-webpack/src/partials/babel.ts
```diff
@@ -6,9 +6,10 @@ const assign = require("lodash/assign");
 const babelLoader = require.resolve("babel-loader");
 
 import { loadXarcOptions } from "../util/load-xarc-options";
-
+const { detectXARCPath } = require("@xarc/app-dev/lib/utils");
 module.exports = function(options) {
   const xarcOptions = loadXarcOptions();
+  const xarcPath = detectXARCPath(xarcOptions.options.cwd);
```
jchip, line 12: as discussed, `detectXARCPath` is just `path.resolve` and is rather extraneous here for example.

packages/xarc-app-dev/src/lib/utils.ts
```diff
@@ -156,3 +156,13 @@ export function detectCSSModule(xOptions) {
   }
```
jchip, line 156: you need to update `loadXarcOptions` because it's using `process.cwd`, also do the same for its copy in `xarc-webpack`

packages/xarc-app-dev/src/lib/utils.ts
```diff
@@ -156,3 +156,13 @@ export function detectCSSModule(xOptions) {
...
   return Boolean(cssModuleSupport);
 }
+
+/**
+ * Detect the xarc path if provided in the app and use it if process.cwd
+ * if xarc cwd is available then check if its absolute or not -
+ * if its aboslute then return it.
+ * if not then create the abosolute path and concatinate with xarc path
+ */
+export function detectXARCPath(xarcCwd: string = process.cwd()) {
```
jchip, line 166: as discussed, this should be removed.

packages/xarc-app-dev/src/config/archetype.ts
durrab: I observed that this is needed because I did the changes to other files in xarc-app-dev and before our xclap.ts there is options.ts which makes a call and it save our xarcOptions.json. After that whenever this method is getting called from xclap.ts and if we pass our user ={cwd:"cwd  value"} then it checks for the cache and cache already exists then its not storing our value and return from there - This is the reason I check for cwd there and save it if its not saved

packages/xarc-app/src/index.ts
durrab: Fixed going to check in

packages/xarc-app-dev/src/config/archetype.ts
```diff
@@ -24,7 +29,7 @@ module.exports = function getDevOptions(user: XarcOptions = {}) {
   // first get legacy configs
   const legacy = getDevArchetypeLegacy();
   // try to read xarc-options.json if it exist and merge it into legacy
-  const xarcOptions = loadXarcOptions();
+  const xarcOptions = loadXarcOptions(process.cwd(), false);
```
jchip, line 32: not `process.cwd`

packages/xarc-app-dev/src/lib/utils.ts
```diff
@@ -90,27 +91,36 @@ export const jsonParser = (key, value) => {
...
 let cachedXarcOptions;
 
-export function loadXarcOptions(dir: string = process.cwd()) {
+export function loadXarcOptions(dir: string = process.cwd(), showError = true) {
+  let cwd;
+  if (process.env.XARC_CWD) {
+    cwd = process.env.XARC_CWD;
+  } else {
+    cwd = dir;
+  }
```
jchip, line 100: line 95-100 can be achieved in line 94 like this:

`dir: string = process.env.XARC_CWD || process.cwd()`

packages/xarc-app-dev/src/lib/utils.ts
```diff
@@ -90,27 +91,36 @@ export const jsonParser = (key, value) => {
...
+      <red>ERROR</>: Electrode xarc fail to load <cyan>.etmp/xarc-options.json</> in
+      dev mode.  This means you are trying to use something not through
+      xarc's development tasks.
+      
+      full path: ${filename}
+      
+      Please run "clap setup-dev" once to initialize the file
+      <cyan>.etmp/xarc-options.json</> before doing your thing that loads
+      xarc's development code.
+      `);
```
jchip, line 121: these extra spaces you added at the beginning of the lines will show in the actual output.

packages/xarc-app/src/index.ts
```diff
@@ -10,16 +10,15 @@ const makeAppMode = require("../lib/app-mode");
 const Path = require("path");
 const constants = require("../lib/constants");
 const logger = require("../lib/logger");
-
 let AppMode;
+const archetype = require("@xarc/app-dev/config/archetype")();
```
jchip, line 14: why is this line moved?

packages/xarc-create-app/test/spec/create-app.spec.js
```diff
@@ -2,20 +2,20 @@ const shcmd = require("shcmd");
...
-  });
+		execSync("rm -rf test/data/create-app-unit-test");
+	});
 
-  it("precondition: directory is created, and create app in cwd", () => {
-    shcmd.exec("echo 'y' | node ../../../src/create-app");
-  });
+	it("precondition: directory is created, and create app in cwd", () => {
+		shcmd.exec("echo 'y' | node ../../../src/create-app");
+	});
```
jchip, line 20: undo these formatting please

packages/xarc-app-dev/src/config/opt2/xarc-options.ts
durrab: Fixed - I just changed the comment for it

packages/xarc-webpack/src/partials/extract-style.ts
```diff
@@ -103,9 +106,10 @@ module.exports = function() {
     const enableShortenCSSNames = xarcOptions.webpack.enableShortenCSSNames;
     const enableShortHash = isProduction && enableShortenCSSNames;
     const localIdentName = `${enableShortHash ? "" : "[name]__[local]___"}[hash:base64:5]`;
+    
 
     return {
-      context: Path.resolve("src"),
+      context: Path.resolve(xarcPath,"src"),
```
jchip: this code was not formatted with prettier.

packages/xarc-app-dev/src/config/archetype.ts
durrab: Fixed

PR #1775, opened 2021-01-05
[fix] CSS module support options and make it work for hapi-app sample

jchip:


packages/xarc-app-dev/src/config/opt2/webpack-options.ts
```diff
@@ -55,11 +55,23 @@ export type WebpackOptions = {
...
+   * Settings:
+   * 1. `true` or "cssOnly" (**default**): CSS module for any file with `.css` extension
+   * 2. RegExp: CSS module for any file that matches the custom RegExp
+   * 3. `"all"`: CSS module for all style files (`css`, `styl`, `sass`, `scss`)
+   * 4. `"byModExt"`: Use the preset RegExp that matches any file ending with `.mod.{ext}` or `.module.{ext}`
+   *    ie: `/\.(mod|module)\.(css|styl|sass|scss)$/i`
+   * 5. `false`: Disable CSS module completely
+   *
+   *
+   * Note: CSS module support require the package `@xarc/opt-postcss`, please add it to your `devDependencies`.
```
divyakarippath, line 69: Now`@xarc/opt-postcss` is not mandatory for css modules. I have removed that check in previous commits.

https://github.com/electrode-io/electrode/pull/1751/files#diff-9fcd63a47bc3b00243ba5c324ad645a1d43f54402bab17ee1e5f2dc8ad239834L61

packages/xarc-app-dev/src/config/opt2/webpack-options.ts
```diff
@@ -55,11 +55,23 @@ export type WebpackOptions = {
...
-   * Set to true to explicitly enable CSS module support for all style files
-   * - **Default: `undefined` (auto detect)**
-   * - If not set, then check env `CSS_MODULE_SUPPORT`
+   * Configure CSS module support.
+   *
+   * Settings:
+   * 1. `true` or "cssOnly" (**default**): CSS module for any file with `.css` extension
+   * 2. RegExp: CSS module for any file that matches the custom RegExp
+   * 3. `"all"`: CSS module for all style files (`css`, `styl`, `sass`, `scss`)
+   * 4. `"byModExt"`: Use the preset RegExp that matches any file ending with `.mod.{ext}` or `.module.{ext}`
```
divyakarippath, line 64: Just wondering if this name will be confusing. Since we are enabling based on file name, should we change this to `byFileName` or something? `byModExt` makes more sense if we let users specify the list of extensions.

jchip: the check is always by filename through the regexp.  in this case, the specific preset regex is to check for the existence of a `mod` or `module` token in the filename.

divyakarippath (approved):

PR #1783, opened 2021-01-12
[MINOR] shared routing for subapps

durrab:


jchip:
update description comments for `options.history` to indicate the new behavior

use `hasOwnProperty` to check if user specified `options.history`

packages/xarc-react-router/src/browser/router-history.ts
```diff
@@ -0,0 +1,4 @@
+import { createBrowserHistory } from "history";
+export default class RouterHistory {
```
jchip, line 2: this is a bit overkill to create a static value, and no `export default`

jchip: also, this will trigger `createBrowserHistory` immediately when app's loaded

durrab: Fixed

packages/xarc-react-router/src/common/index.ts
```diff
@@ -11,9 +11,10 @@ export type ReactRouterFeatureOptions = {
   /**
    * The browser history object - custom browser history object and control which Router to use.
    *
-   * - If it's falsy, then `BrowserRouter` will be used.
-   * - If it's `true`, then `Router` is used with history from `createBrowserHistory` from https://www.npmjs.com/package/history
-   * - Otherwise it's assumed to be a history object and `Router` will be used with it.
+   * - If it's false, then `BrowserRouter` will be used and it will use its own history.
+   * - Note: BrowserRouter use its own history object and its not shared among other subapps
+   * - If it's `true`, then `Router` is used with history from `createBrowserHistory` from https://www.npmjs.com/package/history and the same history object will be shared among all subapps
```
jchip, line 16: what if `history` is not specified?

packages/xarc-react-router/src/browser/react-router-browser.tsx
```diff
@@ -1,9 +1,9 @@
 import { SubAppDef, SubAppFeatureFactory } from "@xarc/subapp";
 import { Router, BrowserRouter } from "react-router-dom";
-import { createBrowserHistory } from "history";
-
+//import RouterHistory from "./router-history"
```
jchip, line 3: remove

packages/xarc-react-router/src/browser/react-router-browser.tsx
```diff
@@ -1,9 +1,9 @@
...
 import { ReactRouterFeatureOptions, _id, _subId } from "../common";
-
+import { createBrowserHistory } from "history";
+const reactRouterHistory = createBrowserHistory();
```
jchip, line 6: only create it when used

samples/subapp2-store/src/subapps/main-body.tsx
```diff
@@ -52,7 +52,7 @@ export const subapp: ReactSubApp = {
     reduxFeature({
```
jchip, line 52: what are these changes in this file for?

packages/xarc-react-router/src/common/index.ts
durrab: I mentioned this here in the last line
* - Otherwise it's assumed to be a history object and `Router` will be used with it. If its undefined then share history object will be used for all subapps

packages/xarc-react-router/src/browser/react-router-browser.tsx
```diff
@@ -20,10 +20,10 @@ export function reactRouterFeature(options: ReactRouterFeatureOptions): SubAppFe
       let history: any;
       let TheRouter: any;
 
-      if (!options.history) {
+      if (options.history === false) {
         TheRouter = BrowserRouter;
       } else {
-        history = options.history === true ? createBrowserHistory() : options.history;
+        history = options.history === true ? reactRouterHistory : (options.hasOwnProperty("history") ? options.history : reactRouterHistory);
```
jchip, line 26: as discussed, the history created in this file is used only if:
1. `options.history` is `true`
2. `options.history` is not `defined`
Otherwise,
1. if it's not `falsy`, assume it's a valid history object and use it with `Router`
2. finally use `BrowserRouter`

jchip: no double `?:` please

samples/subapp2-store/src/subapps/main-body.tsx
durrab: I removed the subapp-web which was not required here - I used it when I was playing with that.

packages/xarc-react-router/src/browser/react-router-browser.tsx
durrab: I have to have reactRouterHistory on top because reactRouterFeature called multiple times for each sub-app and this is a one time call as shared history object

packages/xarc-react-router/src/browser/react-router-browser.tsx
durrab: done

samples/subapp2-store/src/subapps/main-body.tsx
jchip: what does these changes have to do with subapp-web?

PR #1793, opened 2021-01-26
refactored react query unit test, add more test case for browser and node side

zenz34:
refactored react query unit test, add more test case for browser and node side


packages/xarc-react-query/test/spec/browser-index.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+      csrData: {
+        name: "test",
+        getInitialState: () => "test"
+      }
+    });
+
+    expect(res).to.be.an("object");
+
+    const htmlMarkup = renderToString(<res.Component />);
+    expect(htmlMarkup).equal(`<div>test </div>`);
```
jchip, line 70: should there be data after `test`?

PR #1800, opened 2021-01-28
add more tests for react query to increase test coverage

zenz34:
Coverage now:

------------------|---------|----------|---------|---------|-------------------
File              | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
------------------|---------|----------|---------|---------|-------------------
All files         |     100 |    92.31 |   83.33 |     100 |                   
 browser          |     100 |     87.5 |     100 |     100 |                   
  index.tsx       |     100 |     87.5 |     100 |     100 | 52                
 common           |     100 |      100 |       0 |     100 |                   
  feature-info.ts |     100 |      100 |     100 |     100 |                   
  index.ts        |     100 |      100 |       0 |     100 |                   
 node             |     100 |    94.44 |     100 |     100 |                   
  index.tsx       |     100 |    94.44 |     100 |     100 | 76                
------------------|---------|----------|---------|---------|-------------------

=============================== Coverage summary ===============================
Statements   : 100% ( 66/66 )
Branches     : 92.31% ( 24/26 )
Functions    : 83.33% ( 10/12 )
Lines        : 100% ( 64/64 )

### DIFF
--- a/packages/xarc-react-query/src/browser/index.tsx
+++ b/packages/xarc-react-query/src/browser/index.tsx
@@ -46,6 +46,22 @@ export function reactQueryFeature(options: ReactQueryFeatureOptions): SubAppFeat
 
     reactQuery.execute = async function ({ input, csrData }) {
       const dehydratedState = csrData.getInitialState();
+
+      if (
+        (!input.Component && !envHooks.getContainer().get(subAppName)._getExport) ||
+        (!input.Component &&
+          typeof envHooks.getContainer().get(subAppName)._getExport !== "function")
+      ) {
+        console.error("Wrapper component not found!"); // eslint-disable-line
+        return {
+          Component: () => (
+            <div>
+              Error:<p>Wrapper component not found!</p>
+            </div>
+          )
+        };
+      }
+
       const WrapComp = this.wrap({
         Component:
           input.Component || envHooks.getContainer().get(subAppName)._getExport()?.Component,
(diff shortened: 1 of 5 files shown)


packages/xarc-react-query/src/browser/index.tsx
```diff
@@ -46,6 +46,22 @@ export function reactQueryFeature(options: ReactQueryFeatureOptions): SubAppFeat
...
+        return {
+          Component: () => (
+            <div>
+              Error:<p>Wrapper component not found!</p>
+            </div>
+          )
+        };
+      }
+
       const WrapComp = this.wrap({
```
zenz34, line 65: find a potential issue while writing unit test for wrap function call here both at browser and node side.

When input.Component is unset, got an error `TypeError: subapp_1.envHooks.getContainer(...).get(...)._getExport is not a function`
It’s because wrapper component will be gotten through subAppContainer when input.Component is undefined.
Since _getExport function is not required in Typescript declaration, it can be missing. So even we have question mark after that function call   `_getExported()?`, but because it’s undefined, error will be throwed.
Another case is,  if getExported is not a function, it will throw an error as well.

So I added several lines above to avoid the error. It console.error error msg and returns an component with error message.



So I’m adding this code before this.wrap get called. Is it ok as the  error msg? What’s ur suggestion? Thanks inadvance!
      if (
        (!input.Component && !envHooks.getContainer().get(subAppName)._getExport) ||
        (!input.Component &&
          typeof envHooks.getContainer().get(subAppName)._getExport !== "function")
      ) {
        console.error("Wrapper component not found!");
        return {
          Component: <div>Wrapper component not found!</div>
        };
      }

jchip: good catch.  it's not needed though.
`_getExport` is an internal function that the subapp2 engine guarantee to add when user `declareSubApp`.  That's why it has the leading `_`.
when unit test manually create subapp def for mocking test, it should supply the `_getExport` if it's needed.  
it should return a mock subapp module that has the `Component`

PR #1829, opened 2021-02-18
admin server pass thru real terminal columns as env

jchip:


zenz34 (approved):

zenz34:
Is create-app.js supposed to be empty for now? If it is, looks good to me.

PR #1850, opened 2021-05-15
[1.0.50][feat] non-root redux subapp support

gaganjakhotiya:


jchip:
Thanks for the PR.  Can you provide some explanation on the context and use case for the change please?

gaganjakhotiya:
This is to support a similar use-case as `loadSubApp` does [here](https://github.com/electrode-io/electrode/blob/master/packages/subapp-react/src/fe-framework-lib.js#L22). If a `reduxSubApp` needs to be rendered within a composite component we won't be getting the DOM element to render the subapp in.

LANCENET (approved):

PR #1865, opened 2021-11-27
[minor][feat] add custom renderer support for SSR.

ashuverma:
This is to support use cases like css in js e.g. Rendering Styled Components on server

e.g.  in `server` entry (server.js)

```
import {ServerStyleSheet} from "styled-components";
import ReactDOM from "react-dom/server";

export default {
renderer(element, options) {
   const sheet = new ServerStyleSheet();
   try {
      const html = ReactDOMServer.renderToString(.  // This can be renderToNodeStream or Static markup/stream depending on option. 
                              sheet.collectStyles(element)
       );
      const styleTags = sheet.getStyleTags(); 

      return styleTags + html;
    } catch (error) {
        console.log(error);
    }
}

### DIFF
--- a/packages/subapp-react/lib/framework-lib.js
+++ b/packages/subapp-react/lib/framework-lib.js
@@ -59,6 +59,12 @@ class FrameworkLib {
   }
 
   renderTo(element, options) {
+    const { subAppServer } = this.ref;
+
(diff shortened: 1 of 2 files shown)


ashuverma:
@divyakarippath @jchip

divyakarippath (approved):

PR #1870, opened 2022-01-28
Enable redux enhancer

WesleyRen:
This is to make it possible to provide an enhancer when create a Redux store with Electrode.

This is the real change:
https://github.com/electrode-io/electrode/pull/1870/commits/832212b55554e2e7ceab533c6ebcc68dea1623a4

All other files are in the new demo case I created -- `poc-subapp-redux`.
Please let me know if that's be best way to add a demo case.

The demo case created a Redux logger as an enhancer.

Screen Shot 2022-01-27 at 8 10 22 PM

<hr/>

**Not related to the PR code**
I believe CLA signing process is broken.
I have signed, but it says I haven't. And this PR is definitely from my GitHub account. 

Screen Shot 2022-01-27 at 10 18 54 PM


chandub (approved):
LGTM. Thanks

PR #1877, opened 2022-04-20
refactor(CEECORE-2932): replace depricated render methods and unnecessary pacakges

mristic505:
### Description
- Replaced non-static render methods in **frameworkLib** which are either deprecated like `renderToNodeStream ` or limited like `renderToStaticNodeStream` with the new `renderToPipeableNodeStream` method that fully supports **Suspense** and streaming of HTML with “delayed” content blocks https://reactjs.org/docs/react-dom-server.html#rendertopipeablestream.
- Removed `react-async-ssr` package from usage in **frameworkLib** since **SSR** now fully supports **Suspense** via `renderToPipeableNodeStream`.
- Removed `react-async-ssr` from **package.json** files.
- Refactored **SSR-framework test** to work with `renderToPipeableNodeStream` method and added empty comment nodes to expectation results. **All non-static render methods add empty comment nodes around text props including the new `renderToPipeableNodeStream` method**.
- Removed `asyncVerify` package from usage in **SSR-framework test** file and added `getTestWritable` function to call `renderToPipeableNodeStream` method.


packages/subapp-react/lib/framework-lib.js
```diff
@@ -65,29 +64,8 @@ class FrameworkLib {
...
-      if (options.hydrateServerData) {
-        return AsyncReactDOMServer.renderToStringAsync(element);
-      } else {
-        return AsyncReactDOMServer.renderToStaticMarkupAsync(element);
-      }
-    }
-    if (options.hydrateServerData) {
-      return ReactDOMServer.renderToString(element);
+    if (options.suspenseSsr || options.useStream || options.hydrateServerData) {
+      return ReactDOMServer.renderToPipeableStream(element);
```
coston, line 68: I was wondering about opportunity to remove these options, but I think it's valid to leave these as is to simplify the upgrade.

packages/subapp-react/test/spec/ssr-framework.spec.js
```diff
@@ -92,7 +112,8 @@ describe("SSR React framework", function () {
     expect(res).contains("Hello foo bar");
   });
 
-  it("should render Component with stream if enabled", () => {
+  it("should render Component with stream if enabled", async () => {
+    const { writable, output, completed } = getTestWritable();
```
coston, line 116: Clean 🧼

coston (approved):
This is much nicer. Nice job!

PR #1878, opened 2022-05-03
refactor(CEECORE-2932): replace depricated render methods and unnecessary pacakges

mristic505:
### Description
- Replaced non-static render methods in **frameworkLib** which are either deprecated like `renderToNodeStream ` or limited like `renderToStaticNodeStream` with the new `renderToPipeableNodeStream` method that fully supports **Suspense** and streaming of HTML with “delayed” content blocks https://reactjs.org/docs/react-dom-server.html#rendertopipeablestream.
- Removed `react-async-ssr` package from usage in **frameworkLib** since **SSR** now fully supports **Suspense** via `renderToPipeableNodeStream`. Added `getStreamWritable` function to handle `renderToPipeableNodeStream` method.
- Removed `react-async-ssr` from **package.json** files.
- Refactored **SSR-framework test** to work with `renderToPipeableNodeStream` method and added empty comment nodes to expectation results. **All non-static render methods add empty comment nodes around text props including the new `renderToPipeableNodeStream` method**.
- Removed `asyncVerify` package from usage in **SSR-framework test** file.


coston (approved):
Seems to only be updating tests now. Send it 🚀

PR #1883, opened 2022-05-09
refactor(CEECORE-3041): Remove method renderToStaticMarkup from frameworkLib and update test

mristic505:
### DESCRIPTION

- Removed the last remaining static render method `renderToStaticMarkup` due to to it's limited Suspense support and official React's recommendation.
- Changed the original implementation of `renderToPipeableStream` and made it writable within the frameworkLib file.
- Updated the test accordingly.


coston (approved):
This looks good. I've validated that this is working with an existing sample, `poc-subapp`

PR #1887, opened 2022-05-24
major(CEECORE-3087): Upgrade xarc-react-router package to V6

mristic505:
### Description
- Upgraded xarc-react-router package.json to current dependencies. 
- Refactored `react-router-node.tsx` and `react-router-browser.tsx` files to work with `react-router` V6.

### Key Changes
- React Router v6 uses `Routes` component instead of `Switch`. 
- `Router` does not support `history` prop anymore. We replaced it with `HistoryRouter` for that reason.
- `StaticRouter` does not support context prop anymore.

### How to Test

- Navigate to `packages/xarc-react-router/src/` directory
- Run `fyn && fun text`

### Key Notes

`unstable_HistoryRouter as HistoryRouter` is used here instead of `<Router>` because react_router v6 removed history prop support for `<Router>`. This API is currently prefixed as unstable_ because you may unintentionally  add two versions of the history library, the one you have added  to the package.json and whatever version React Router uses internally. Once react-router has a mechanism to detect mis-matched versions, this API will remove its unstable_ prefix. [https://reactrouter.com/docs/en/v6/routers/history-router](https://reactrouter.com/docs/en/v6/routers/history-router). 
For now, use the unstable_HistoryRouter version here.


smuthya (approved):
LGTM

PR #1888, opened 2022-06-02
fix: update dependencies to resolve build issues

smuthya:
With typescript version 4.7.2, tests are failing in webpack-config-composer, xarc-app-dev, xarc-subapp packages. Updated the version of ts-node to fix this.

data-reactroot isn't getting created due to Context.Provider, so removed it from xarc-react test case

enzyme breaks with cheerio version v1.0.0-rc.11, so had to set cheerio to =1.0.0-rc.10 version in samples/stylus-sample, samples/react-jest-app to resolve this issue

Reference :
https://github.com/facebook/jest/issues/12655
https://github.com/facebook/react/issues/15012
https://github.com/enzymejs/enzyme/issues/2558

### DIFF
--- a/packages/webpack-config-composer/package.json
+++ b/packages/webpack-config-composer/package.json
@@ -55,7 +55,7 @@
     "sinon": "^9.2.2",
     "sinon-chai": "^3.5.0",
     "source-map-support": "^0.5.19",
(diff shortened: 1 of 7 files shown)


mristic505:
LGTM

PR #1894, opened 2022-06-24
fix: failed tests and checks in subapp-react

mristic505:
- fix failed checks in subapp-react
- fix failed tests for undefined props in subapp-react


smuthya (approved):
LGTM

PR #1909, opened 2022-10-17
fix lint issues in sample apps

smuthya:
Fixed lint issues in stylus-sample and react-jest-app sample applications


chandub (approved):
LGTM

PR #1915, opened 2022-11-21
fix issues related to hmr, isomorphic loading, error reporting

ashuverma:
- Don't inject hot-update bundles in the template to avoid below error:

Screenshot 2022-11-22 at 12 06 13 AM

- Add `crossorigin="anonymous"` to our scripts for error reporting. By default, error message will include only `Script Error.`
- Skip container entry to avoid errors while refreshing apps on server
Screenshot 2022-11-21 at 11 55 39 PM

- Handle isomorphic loading e.g. images, css modules etc. during first load, the app will start before webpack finish compiling, so webpack will inform app that modules refreshed when they didn't. sets a one time flag to avoid refreshing when first start up, reset this one time flag only after isomorphic initialise is set.

### DIFF
--- a/packages/subapp-web/lib/load.js
+++ b/packages/subapp-web/lib/load.js
@@ -111,7 +111,7 @@ Response: ${err || body}`
         // if we have to inline the subapp's JS bundle, we load it for production mode
         const src = Fs.readFileSync(Path.resolve("dist/js", bundleAsset.name)).toString();
         const ext = Path.extname(bundleAsset.name);
-        if (ext === ".js") {
+        if (ext === ".js" && !src.endsWith(".hot-update.js")) {
           inlineSubAppJs = `<script>/*${name}*/${src}</script>`;
         } else if (ext === ".css") {
           inlineSubAppJs = `<style id="${name}">${src}</style>`;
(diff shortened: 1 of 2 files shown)


packages/subapp-web/lib/load.js
```diff
@@ -156,11 +156,11 @@ Response: ${err || body}`
               .concat(cdnJsBundles[ep])
               .reduce((a, jsBundle) => {
                 const ext = Path.extname(jsBundle);
-                if (ext === ".js") {
+                if (ext === ".js" && !jsBundle.endsWith(".hot-update.js")) {
                   if (context.user.headEntries) {
                     headSplits.push(`<link rel="preload" href="${jsBundle}" as="script">`);
                   }
-                  a.push(`<script src="${jsBundle}" async></script>`);
+                  a.push(`<script src="${jsBundle}" async crossorigin="anonymous"></script>`);
```
arunvishnun, line 163: Hi @ashuverma Thanks for changes. 
Looks like this changes is causing a test failure for `should load bundles for the subapp` -  ``Error: Timeout of 2000ms exceeded. ``. Do you know what is going wrong ?

ashuverma: Any examples where I can see this?

PR #1936, opened 2023-01-17
feat(node-sass): update node-sass and sass-loader to support node v16

shubham2811:
Update node-sass and sass-loader dependencies to support node v16 installation for the @xarc/opt-sass package


arunvishnun (approved):

smuthya (approved):
LGTM

PR #1962, opened 2023-05-04
fix(subapp-server): :ambulance: Fix memory leak in template cache

ashuverma:
Remove dependence on request url and use configured route path.


arunvishnun (approved):
Looks good to me, @ashuverma could you push change files as well (`rush change`) so that we can publish

PR #2002, opened 2024-01-05
Downgrade jsdom to v21 for backward node version compatibility

shubham2811:
jsom v23(latest as of now) require minimum node 18 version which is breaking ci tests in case of node v14.So downgrading it to v21


shubham2811:
Dropping node 14 support.Hence closing PR

PR #2005, opened 2024-02-28
Csp directive

RahulTinku:
Add ablity for applications to set their CSP headers which may include style-src & script-src. Applications can set their nonce value or use default nonce value provided by electrode as usual. 

getCSPHeader is added in route options, which accepts a function. In this function application can generate the CSP and return it as string.

### DIFF
--- a/packages/subapp-server/lib/fastify-plugin.js
+++ b/packages/subapp-server/lib/fastify-plugin.js
@@ -39,14 +39,14 @@ function makeRouteHandler({ path, routeRenderer, routeOptions }) {
 
   return async (request, reply) => {
     try {
-      const { styleNonce = "", scriptNonce = ""} = setCSPNonce({ routeOptions });
+      const { styleNonce = "", scriptNonce = "" } = setCSPNonce({ routeOptions });
 
       // wait for webpack stats to be valid if webpackDev
       if (webpackDev) {
         await until(() => request.app.webpackDev.valid === true, 400);
         console.log(`Webpack stats valid: ${request.app.webpackDev.valid}`);
       }
-      
+
       const context = await routeRenderer({
         useStream,
         mode: "",
@@ -58,11 +58,11 @@ function makeRouteHandler({ path, routeRenderer, routeOptions }) {
 
       let cspHeader;
       /** If csp headers are provided by application in route options then use that otherwise generate CSP headers */
-      if(routeOptions.cspHeaderValues instanceof Function ){
+      if (routeOptions.cspHeaderValues instanceof Function) {
         const rawCSPHeader = routeOptions.cspHeaderValues({ styleNonce, scriptNonce });
         // Replace newline characters and spaces
         cspHeader = rawCSPHeader.replace(/\s{2,}/g, " ").trim();
-      }else{
+      } else {
         cspHeader = getCSPHeader({ styleNonce, scriptNonce });
       }
       if (cspHeader) {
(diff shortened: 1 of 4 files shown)


packages/subapp-server/lib/fastify-plugin.js
```diff
@@ -56,8 +56,15 @@ function makeRouteHandler({ path, routeRenderer, routeOptions }) {
       const data = context.result;
       const status = data.status;
 
-      const cspHeader = getCSPHeader({ styleNonce, scriptNonce });
-
+      let cspHeader;
+      /** If csp headers are provided by application in route options then use that otherwise generate CSP headers */
+      if(routeOptions.getCSPHeader && typeof routeOptions.getCSPHeader === "function"){
+        const rawCSPHeader = routeOptions.getCSPHeader({ styleNonce, scriptNonce });
```
arunvishnun, line 62: Since this is to set CSP header, I guess a better suitable function name would be something like `setCSPHeader` instead of `getCSP*` in routeOptions. (Corresponding change in route as well)

arunvishnun:
LGTM in general, seems there is lint issues (spacings)

packages/subapp-server/lib/fastify-plugin.js
divyakarippath: Personally I prefer getCSPHeader or may be just `cspHeaderValues`, since the user is not really setting the header, they will be only returning the list of header values (similar to getCSPHeader function in our package).

divyakarippath: @arunvishnun @RahulTinku  what are your thoughts on  `cspHeaderValues` as the option name?

jchip:
Screenshot 2024-02-28 at 2 13 45 PM


from ChatGPT.

packages/subapp-server/lib/fastify-plugin.js
annelin07: Agree, or probably `getCSPDirectives`.

samples/poc-subappv1-csp/src/routes.js
```diff
@@ -30,7 +48,8 @@ export default {
     // Enable one of these to use CSP header
     cspNonce: true,
```
arunvishnun, line 49: need to disable this for this sample since app code is passing down header info

RahulTinku: cspNonce value would be required, as if user wants to set script-src or style-src, and doesnot set the nonce, we'd generate and pass as params in `cspHeaderValues` in routes. This is to keep the as-is behaviour.

packages/subapp-server/lib/fastify-plugin.js
RahulTinku: I agree with `cspHeaderValues` as we are just adding the values. Setting the headers would be taken care by subapp-server.

RahulTinku:
Thank you @jchip, I shall update the code accordingly.

divyakarippath (approved):

samples/poc-subappv1-csp/src/routes.js
arunvishnun: ah ok, my bad. nvm

arunvishnun:
@RahulTinku Could you create [change files](https://github.com/electrode-io/electrode/blob/master/CONTRIBUTING.md#developer-flow).  
Also, seems there is lint issue [here](https://github.com/electrode-io/electrode/pull/2005/files#diff-47234878f73deff70e5dc4543ed77bb21f26299ce8f4bfb6ae5b2de0b5f4befbR61) - Spacings are not consistent with other parts of the file.

arunvishnun (approved):

RahulTinku:
Hi @arunvishnun , I have added changelog. 
Thanks

PR #2017, opened 2024-07-03
fix: windows path issue

GauravKumarSingh:
"npm run lint" throwing error in windows.
"Path.posix.join" combine path according to unix system only (i.e  forward slash)
It is not working in windows. After replacing it with "Path.join" It is working fine.


kylejeske (approved):
lgtm

PR #2040, opened 2025-02-17
Update react router dom to v7

shubham2811:
Replace react-router-dom v6 with [react-router v7](https://reactrouter.com/upgrading/v6)


arunvishnun (approved):
LGTM

PR #2048, opened 2025-06-09
React 19

RahulTinku:
Xarc-react , recoil pacakge update to react 19


packages/xarc-react-recoil/package.json
```diff
@@ -25,19 +25,19 @@
...
     "@babel/eslint-parser": "^7.25.1",
     "@istanbuljs/nyc-config-typescript": "^1.0.1",
-    "@testing-library/react": "^11.2.5",
+    "@testing-library/react": "^16.3.0",
     "@types/chai": "^4.2.14",
     "@types/mocha": "^10.0.1",
     "@types/node": "^20.14.10",
-    "@types/react": "^18",
-    "@types/react-dom": "^18",
+    "@types/react": ">=19",
```
arunvishnun, line 39: Better to keep it `"^19.1.0"`

packages/xarc-react/src/browser/subapp-inline-component.tsx
```diff
@@ -1,47 +1,35 @@
...
+    xarcV2.debug("SubAppInlineComponent subapp", subapp.name, inlineId);
+  }, [subapp, inlineId]);
+  // Pass all props to SubAppStartComponent
+  return (
+    <SubAppStartComponent
+      subapp={subapp}
+      pipeline={pipeline}
+      __props={__props}
+    />
+  );
```
arunvishnun, line 34: In the current code `shouldComponentUpdate`, seems to prevent a re render.
But our  new functional component may not prevent such re-renders the way `shouldComponentUpdate()` did. I Ihink we would need to wrap SubAppInlineComponent in React.memo. Otherwise, it may re-render more frequently than before.

```js
function SubAppInlineComponent(props: SubAppInlineComponentPropsType): React.JSX.Element {
  // ....
}

export const MemoizedSubAppInlineComponent = React.memo(SubAppInlineComponent, () => true);````

Comment (RahulTinku) [reply packages/xarc-react/src/browser/subapp-inline-component.tsx]: updated implementation and added logic for checking module.

File: packages/xarc-react/src/browser/subapp-as-component.tsx [hunk 3]
```diff
@@ -15,57 +15,26 @@ import { SubAppInlineOptions } from "../common";
...
-  // subapp.id -> unique definition declare time id
-  // subapp.asCompId -> incrementing index of as component instance
-  // subapp.asCompId resets to zero when subapp is re-declared for HMR
-  //
-  // the prepare step needs hint from user on how many pipeline is needed for a subapp
-  // to be used as nesting comp
-  // when SubAppStartComponent needs a pipeline to use, it just look up as-comp pipeline
-  // with count index same as asCompId.
-  //
-
```
arunvishnun, line 58: Lets add the comments back

RahulTinku:
Changes merged with different PR https://github.com/electrode-io/electrode/pull/2055

